PR #3, opened 2016-10-15
Lights Board Code

enochtsang:
This is my first embedded systems project, please be gentle :)
Overall I've opted for verbosity over efficiency.

Most of the files contain initialization code which have been generated by STM32CubeMX, for the most part I did not edit these files. Only `main.c` was heavily edited by me. 

I did have to make one change to library files:
`stm32f4xx_hal_can.h` lines 412 and 416 were added by me. They were missing a define for an interrupt flag `CAN_FLAG_FMP0` and `CAN_FLAG_FMP1`. I use this define in `HAL_CAN_RxCpltCallback` in `Lights.c`.

CAN is all tested via loopback mode. So the receiving filter works, and sending also works. Updating lights is not tested at all, awaiting test board completion.

**Only `main.c`, `Lights.h,` and `Lights.c` need to be reviewed**  

NOTE: For `main.c`, the functions `SystemClock_Config()` and `MX_GPIO_Init()` are completely generated, I wouldn't review those, otherwise the rest of `main.c` is good to review.

---

This change is [Reviewable](https://reviewable.io/reviews/ucsolarcarteam/epsilon-embedded-software/3)


lahunter:
In general - maybe don't put 31 commits up for review at once, there's no way anyone's going to be able to catch everything in this. WRT your verbosity vs efficiency comment - you're working with a very fast piece of embedded hardware. You should be able to do pretty much everything in as inefficient manner as possible and not have any problems, so you can start off being super verbose, and optimize later if the code DOES turn out to be too inefficient.  Also, I'd make sure that all of the code is in the allowable regions in main.c - if it's somewhere else, regenerating will wipe your code. I can't tell if it's not,  but make sure.

---

Review status: 0 of 116 files reviewed at latest revision, 8 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU8F98UBSrMPtexE1Z9:-KU9CNykPoIRwo7Py2pZ:bh46v3s) (no related file):*
<details><summary><i>Previously, enochtsang (Enoch Tsang) wrote…</i></summary>
> Have I been too liberal with my use of globals? I heard it's a common thing in C. I didn't want to put the effort of multiple arguments in the void* arg if it wasn't necessary.
</details>
I'm fine with globals, but for anything not needed between different files, you should make them static, and for things needed between files, declare them extern in a header.

---

*[a discussion](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU8FHFN4Rn38mqHQJ65:-KU9DHIcYhDEw_IyMvve:b1wzr3) (no related file):*
<details><summary><i>Previously, enochtsang (Enoch Tsang) wrote…</i></summary>
> Are my stack sizes too small on my threads?
</details>
Try with the defaults first as you've done, and then you'll want to do some profiling at some point to see if they are way to large or way too small  - there's not really any way to do that otherwise

---

*[EpsilonLights/Inc/Lights.h, line 39 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9Gp9VOypPf4U3vjN4:-KU9Gp9VOypPf4U3vjN5:b-6p59pf) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Inc/Lights.h#L39)):*
> ```objc
>     struct
>     {
>         unsigned int lowBeams : 1;
> ```

Just a warning that while this is really useful, writing to the bitfield and reading from the uint8 (or vice versa) is technically UB, so might not work properly under some compilers. GCC is fine with it, but just a note for future debugging. 

---

*[EpsilonLights/Src/Lights.c, line 23 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9IZoi3c90HKfecEGO:-KU9IZojs7CJEwoVPrFK:b-993czr) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Src/Lights.c#L23)):*
> ```C
> 
>     // NOTE: All Lights Out pins are active low
>     for (;;)
> ```

I feel like you should be launching this as a scheduled task on the thread instead of an infinite loop, but I might be wrong.

---

*[EpsilonLights/Src/Lights.c, line 91 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9L0IM_RwD_5tjXnrU:-KU9L0INrerl9Ciz8RuB:b-8a05gs) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Src/Lights.c#L91)):*
> ```C
>     uint8_t prevSigState = 0;
> 
>     for (;;)
> ```

As above

---

*[EpsilonLights/Src/Lights.c, line 151 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9L1_IV0ddePxpqzXL:-KU9L1_IV0ddePxpqzXM:b-8rlkgc) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Src/Lights.c#L151)):*
> ```C
>     uint32_t prevWakeTime = osKernelSysTick();
> 
>     for (;;)
> ```

As Above

---

*[EpsilonLights/Src/Lights.c, line 191 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9L2wRPZ2JD3FfN0bn:-KU9L2wRPZ2JD3FfN0bo:b-8a05gs) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Src/Lights.c#L191)):*
> ```C
>     uint32_t prevWakeTime = osKernelSysTick();
> 
>     for (;;)
> ```

As above

---

*[EpsilonLights/Src/main.c, line 122 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9BOYCgGJsA_84ADwq:-KU9BOYCgGJsA_84ADwr:bepdiqi) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Src/main.c#L122)):*
> ```C
>     osThreadDef(blinkLightsTask, blinkSignalLights, osPriorityNormal, 1, configMINIMAL_STACK_SIZE);
>     blinkLightsHandle = osThreadCreate(osThread(blinkLightsTask), NULL);
>     /* USER CODE END RTOS_THREADS */
> ```

For readability, might be good to break this up a little bit - kind of a block of only semi-related, unaligned code with a lot of arguments.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-:-KU9EPtc6ofHdpB2NEIr:bgp40y7)*

lahunter:
Reviewed 2 of 113 files at r1, 2 of 111 files at r3, 99 of 109 files at r4, 13 of 14 files at r5.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3)*

enochtsang:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU8F98UBSrMPtexE1Z9:-KU9QsQCXneZZs127HSN:b-g1u8uv) (no related file):*
<details><summary><i>Previously, lahunter wrote…</i></summary>
> I'm fine with globals, but for anything not needed between different files, you should make them static, and for things needed between files, declare them extern in a header.
</details>
Well noted, will change.

---

*[a discussion](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU8FHFN4Rn38mqHQJ65:-KU9QvpEkTkb5qdLD2_T:bad65uc) (no related file):*
<details><summary><i>Previously, lahunter wrote…</i></summary>
> Try with the defaults first as you've done, and then you'll want to do some profiling at some point to see if they are way to large or way too small  - there's not really any way to do that otherwise
</details>
I think if the program is crashing then we'll know if it's too small, I did have them set to 0 at first and saw many funky errors, I don't see those kinds of errors now so I think it's ok. Looks like there's no harm in having them way too big either, so if you have no other comments I think we'll just keep as is.

---

*[EpsilonLights/Inc/Lights.h, line 39 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9Gp9VOypPf4U3vjN4:-KU9Qqi-bTs3tXh1mfp9:b-rsekm0) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Inc/Lights.h#L39)):*
<details><summary><i>Previously, lahunter wrote…</i></summary>
> Just a warning that while this is really useful, writing to the bitfield and reading from the uint8 (or vice versa) is technically UB, so might not work properly under some compilers. GCC is fine with it, but just a note for future debugging. 
</details>
What does UB mean

---

*[EpsilonLights/Src/Lights.c, line 23 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9IZoi3c90HKfecEGO:-KU9QVbr5LNXAP09g3aS:b-dg7ll8) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Src/Lights.c#L23)):*
<details><summary><i>Previously, lahunter wrote…</i></summary>
> I feel like you should be launching this as a scheduled task on the thread instead of an infinite loop, but I might be wrong.
</details>
In FreeRTOS it looks like using osDelayUntil is the proper way of making a scheduled task. Maybe I'm wrong, but I didn't see anything in the libraries for a scheduled task.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3)*

lahunter:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU8FHFN4Rn38mqHQJ65:-KU9TDGGPONHJK7mIuYD:b-9np5bg) (no related file):*
<details><summary><i>Previously, enochtsang (Enoch Tsang) wrote…</i></summary>
> I think if the program is crashing then we'll know if it's too small, I did have them set to 0 at first and saw many funky errors, I don't see those kinds of errors now so I think it's ok. Looks like there's no harm in having them way too big either, so if you have no other comments I think we'll just keep as is.
</details>
Well, the harm in making them way too big is you would run out of memory eventually, and much sooner than you'd think. But if they're not causing any problems right now, you're good for right now.

---

*[EpsilonLights/Inc/Lights.h, line 39 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9Gp9VOypPf4U3vjN4:-KU9SZ6GjNFXMEBbXbGE:b-mpnduh) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Inc/Lights.h#L39)):*
<details><summary><i>Previously, enochtsang (Enoch Tsang) wrote…</i></summary>
> What does UB mean
</details>
Undefined Behavior

---

*[EpsilonLights/Src/Lights.c, line 23 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9IZoi3c90HKfecEGO:-KU9T-GPdvveMJ6THvNt:bsvt5l8) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Src/Lights.c#L23)):*
<details><summary><i>Previously, enochtsang (Enoch Tsang) wrote…</i></summary>
> In FreeRTOS it looks like using osDelayUntil is the proper way of making a scheduled task. Maybe I'm wrong, but I didn't see anything in the libraries for a scheduled task.
</details>
Ok, feels pretty janky but if that's the accepted method, cool.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3)*

enochtsang:
Will do, 1

---

Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[EpsilonLights/Inc/Lights.h, line 39 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9Gp9VOypPf4U3vjN4:-KUEoO96rbMxJdXgUatA:bmmzqra) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Inc/Lights.h#L39)):*
<details><summary><i>Previously, lahunter wrote…</i></summary>
> Undefined Behavior
</details>
Will make a note as a comment.

---

*[EpsilonLights/Src/Lights.c, line 151 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9L1_IV0ddePxpqzXL:-KUEoTf2p4s-e6NB6Pbo:b-896fix) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Src/Lights.c#L151)):*
<details><summary><i>Previously, lahunter wrote…</i></summary>
> As Above
</details>
Done.

---

*[EpsilonLights/Src/Lights.c, line 191 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9L2wRPZ2JD3FfN0bn:-KUEoUrbG6KEyprl-gE-:b-896fix) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Src/Lights.c#L191)):*
<details><summary><i>Previously, lahunter wrote…</i></summary>
> As above
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-:-KU9SdNb8kaINUunFLJK:b-8pq23b)*

bill-luu:
Reviewed 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[EpsilonLights/Src/Lights.c, line 95 at r6](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KUFLft4Y6bwRCnH6gPY:-KUFLftBFpG7gEG4FOXj:bripnrm) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/e7726cd0e8c0fd52dbfd45c19ae94f6da0164325/EpsilonLights/Src/Lights.c#L95)):*
> ```C
>         osDelayUntil(&prevWakeTime, LIGHTS_UPDATE_FREQ);
> 
>         // Check if both signa lights inputs are DISABLED
> ```

Typo on signal

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3)*

JordanHeinrichs:
Reviewed 1 of 14 files at r5.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[EpsilonLights/Inc/Lights.h, line 39 at r5](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KU9Gp9VOypPf4U3vjN4:-KUQ5EaG_lsibD2kb3Ld:b-tx16df) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/2681c1737977b0065ed71a38452c1e0260850333/EpsilonLights/Inc/Lights.h#L39)):*
<details><summary><i>Previously, enochtsang (Enoch Tsang) wrote…</i></summary>
> Will make a note as a comment.
</details>
Also note that it is not defined between versions of GCC either. As shown in http://stackoverflow.com/a/1490144                 

---

*[EpsilonLights/Inc/Lights.h, line 65 at r6](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KUQ8ZZuDv4TwNAS0Oc7:-KUQ8ZZuDv4TwNAS0Oc8:b-gw6yin) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/e7726cd0e8c0fd52dbfd45c19ae94f6da0164325/EpsilonLights/Inc/Lights.h#L65)):*
> ```objc
> // Task for sending lights statuses via CAN
> // arg : osMutexId* canHandleMutex
> void reportLightsToCan(void const* arg);
> ```

For both this and sendHeartbeat I would recommend renaming them to something that makes it clear that it is a thread/task. From just looking at the name it looks like it should just send it and return.

---

*[EpsilonLights/Src/Lights.c, line 155 at r6](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KUQAYAOUOSCfTZKFYkS:-KUQAYAOUOSCfTZKFYkT:bfxdp9b) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/e7726cd0e8c0fd52dbfd45c19ae94f6da0164325/EpsilonLights/Src/Lights.c#L155)):*
> ```C
>         osDelayUntil(&prevWakeTime, LIGHTS_STATUS_FREQ);
>         // Toggle blue LED for every CAN message sent
>         HAL_GPIO_TogglePin(LED_BLUE_GPIO_Port, LED_BLUE_Pin);
> ```

These comments apply to this function and the function below (sendHeartbeat). 
This LED won't blink every time CAN is sent. It is theoretically possible that can never be sent yet the LED still be blinking away. What I would recommend is turning it within the critical section and make the above line just turn it off. This way it will only be on for while it is sending until the next wake up. 

---

*[EpsilonLights/Src/Lights.c, line 157 at r6](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KUQBT1tZUgmRwVGdAfd:-KUQBT1tZUgmRwVGdAfe:bc8k3fm) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/e7726cd0e8c0fd52dbfd45c19ae94f6da0164325/EpsilonLights/Src/Lights.c#L157)):*
> ```C
>         HAL_GPIO_TogglePin(LED_BLUE_GPIO_Port, LED_BLUE_Pin);
> 
>         if (osMutexWait(canHandleMutex, 0) != osOK)
> ```

While may work in practice it does have a very big risk associated to it.  The risk is that it will not enter the critical section as the other thread could be in it. Instead of waiting it will just uselessly blink the LED misleading the person debugging into thinking the message is actually being sent. Even worse is even if it is only missed a few times for something as important as a heartbeat the time between messages is extremely important it could result in the receiving end thinking the lights board is not functioning.

There are a few approaches to solving this: 
 - One is that you could have only one CAN sending thread, which runs at the greatest common denominator of the few periods and uses counters to get the timing right. 
 - The other options if you are content with only have two CAN threads is to implement something like the Peterson's Solution.

---

*[EpsilonLights/Src/Lights.c, line 177 at r6](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KUQ8I90kyK8wXWw-c1l:-KUQ8I90kyK8wXWw-c1m:beswlop) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/e7726cd0e8c0fd52dbfd45c19ae94f6da0164325/EpsilonLights/Src/Lights.c#L177)):*
> ```C
>         HAL_CAN_Transmit_IT(&hcan2);
> 
>         if (osMutexRelease(canHandleMutex) != osOK)
> ```

There really isn't any use of having the continue on error here. The continue will not do anything as it is at the end of the loop anyway. If the mutex in this case fails to release then stuff has gone really wrong

---

*[EpsilonLights/Src/Lights.c, line 209 at r6](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3#-KUQ7Ak2qUs1qS2Lhjzn:-KUQ7Ak2qUs1qS2Lhjzo:b-25tulu) ([raw file](https://github.com/ucsolarcarteam/epsilon-embedded-software/blob/e7726cd0e8c0fd52dbfd45c19ae94f6da0164325/EpsilonLights/Src/Lights.c#L209)):*
> ```C
>         HAL_CAN_Transmit_IT(&hcan2);
> 
>         if (osMutexRelease(canHandleMutex) != osOK)
> ```

Same as above.

There really isn't any use of having the continue on error here. The continue will not do anything as it is at the end of the loop anyway. If the mutex in this case fails to release then stuff has gone really wrong

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ucsolarcarteam/epsilon-embedded-software/3)*

PR #5, opened 2016-11-23
Driver controls

enochtsang:
Driver Controls needs petersons, and running average for analog inputs.


EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -42,15 +57,24 @@ void sendMusicTask(void const* arg)
 {
```
JordanHeinrichs: In the future instead of having multiple tasks sending CAN I think just having one task would be so much simpler. In that  case you would have absolutely no need for mutexs or petersons or anything like that. There is a tiny bit of complexity getting each send function to go in the correct frequency but simply do:
```cpp
 int counter = 4; // Obviously have clearer names and value defines
 for (;;)
    osDelayUntil(&prevWakeTime, FASTEST_INTERVAL); // in this case 50
   if (counter == 0)
   {
      sendMusic();
      counter = 4;
   }
   sendDriverTask();
}
```
If it is done like this there is no risk of deadlock and very quick and easy to add new tasks;

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -60,15 +84,28 @@ void sendDriverTask(void const* arg)
...
+
+        if (osMutexWait(canHandleMutex, 0) != osOK)
         {
             continue;
         }
+
         hcan2.pTxMsg->StdId = DRIVER_STDID;
-        hcan2.pTxMsg->Data[0] = 1;
+        hcan2.pTxMsg->DLC = 4;
+        hcan2.pTxMsg->Data[0] = 0;
```
JordanHeinrichs: `std::memset(hcan2.pTxMsg->Data, 0, 4);`

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -100,6 +102,21 @@ void sendDriverTask(void const* arg)
         hcan2.pTxMsg->Data[1] = 0;
         hcan2.pTxMsg->Data[2] = 0;
         hcan2.pTxMsg->Data[3] = 0;
+
+        if (HAL_ADC_PollForConversion(&hadc1, 1000000) == HAL_OK)
```
JordanHeinrichs: So if it is unable to read, so send last value? I am unsure of that with the safety implications of having an input stuck high. But setting 0 could be lots of change of some read and some don't but overall I think it would be safer (ideally set 0 into running average).

I think this might need a comment highlighting what is happening here.

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -100,6 +102,21 @@ void sendDriverTask(void const* arg)
...
+        {
+            regen_ADCValue = HAL_ADC_GetValue(&hadc1);
+        }
+        hcan2.pTxMsg->Data[0] += (regen_ADCValue & 0x000000ffUL);
```
JordanHeinrichs: I think `=` would be better here and below use `|=` instead of `+=`. They most likely could be the same but `|=` is more explicit of what you are doing (and if it is somehow signed there can be no overflow).

EpsilonDriverControls/Inc/DriverControls.h
```diff
@@ -7,15 +7,34 @@
...
+#define DRIVE_COMMANDS_FREQ 100
+#define MOTOR_DRIVE_STDID 0x501U
+#define MOTOR_DRIVE_DLC 8
+#define MOTOR_POWER_STDID 0x502U
+#define MOTOR_POWER_DLC 8
+#define MOTOR_RESET_STDID 0x503U
+
+#define ADC_POLL_TIMEOUT 1000
+
+#define MAX_FORWARD_RPM 1000 //Was 1298 on old system
```
JordanHeinrichs: `//Was 1298 on old system` I think this comment does not need to make it over. What first gen CCS did is no longer relevant =P

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -17,7 +19,7 @@ void sendHeartbeatTask(void const* arg)
         }
 
         hcan2.pTxMsg->StdId = HEARTBEAT_STDID;
-        hcan2.pTxMsg->DLC = 1;
+        hcan2.pTxMsg->DLC = HEARTBEAT_DLC;
```
JordanHeinrichs: 👍  with swapping it out to constants (or silly c lang defines)

JordanHeinrichs:
hmm, a lot of them got "outdated" but it was just format changes and I think most of them still are valid.

EpsilonDriverControls/Src/DriverControls.c
enochtsang: Actually this is a bug, I didn't realize it would send the last value

JordanHeinrichs:
LGTM 👍

PR #7, opened 2017-01-21
fixed astyle url

badeniran:


enochtsang:
Baba please run the format script so we can merge this.

PR #9, opened 2017-01-26
Ccs can config

enochtsang:
Link to docs for reference

https://docs.google.com/spreadsheets/d/1soVLjeD9Sl7z7Z6cYMyn1fmn-cG7tx_pfFDsvgkCqMU/edit#gid=0


EpsilonCCS/Src/main.c
```diff
@@ -357,29 +354,26 @@ static void MX_CAN1_UserInit(void)
     static CanRxMsgTypeDef rxMessage;
     hcan1.pTxMsg = &txMessage;
     hcan1.pRxMsg = &rxMessage;
-    // hcan1.pTxMsg->StdId = 0x0; // CAN message address, set in Lights.c
+    // hcan1.pTxMsg->StdId = 0x0; // CAN message address
     hcan1.pTxMsg->ExtId = 0x0; // Only used if (hcan1.pTxMsg->IDE == CAN_ID_EXT)
     hcan1.pTxMsg->RTR = CAN_RTR_DATA; // Data request, not remote request
     hcan1.pTxMsg->IDE = CAN_ID_STD; // Standard CAN, not Extended
-    hcan1.pTxMsg->DLC = 1; // Data size in bytes
+    // hcan2.pTxMsg->DLC = ; // Value is variable
```
JordanHeinrichs: Within the MBED code it was not variable, everything was 8 bytes always. I am pretty sure all Tritium stuff is a constant 8 bytes as well.  Within the "UofC Solar Car Team Communications Protocol" I don't see anything indicating non-standard CAN byte sizes, if they do exist then the document should be updated for that.

Same comment for CAN2 DLC

EpsilonCCS/Src/main.c
```diff
@@ -357,29 +354,26 @@ static void MX_CAN1_UserInit(void)
...
-    // sFilterConfig.FilterIdHigh = 0; // Filter registers need to be shifted left 5 bits
-    // sFilterConfig.FilterIdLow = 0; // Filter registers need to be shifted left 5 bits
-    sFilterConfig.FilterMaskIdHigh = DRIVERS_INPUTS_STDID << 5; // Unused
-    sFilterConfig.FilterMaskIdLow = 0; // Unused
+    sFilterConfig.FilterIdHigh = 0; // Accept All
+    sFilterConfig.FilterIdLow = 0; // Accept
+    sFilterConfig.FilterMaskIdHigh = 0; // Accept All
+    sFilterConfig.FilterMaskIdLow = 0; // Accept All
     sFilterConfig.FilterFIFOAssignment = 0;
     sFilterConfig.FilterActivation = ENABLE;
```
JordanHeinrichs: I appears like the filter is just setup to accept all. Why not disable the FilterActivation in this case?

enochtsang: You've worked with the STM32 before, this works as expected?

EpsilonCCS/Src/main.c
enochtsang: Well all the tritium stuff uses the maximum amount of bytes, some of the internal ones that I've listed do not. We are approaching max CAN bus capacity so I feel like it's a good idea to make it smaller when possible.

JordanHeinrichs:
Looked through all the changes (even though already merged). It all looks good 👍

PR #12, opened 2017-03-11
add code to initalize data structures

enochtsang:
Sorry the original data structures to master, might be easier to review this change right on the repo.


EpsilonCCS/Inc/Data/KeyMotorData.h
```diff
@@ -19,4 +18,6 @@ struct KeyMotorData
     float m1BusCurrent;
     float m1BusVoltage;
     float m1VehicleVelocity;
-}
\ No newline at end of file
+}
```
bill-luu: Semicolon?

EpsilonCCS/Inc/Data/DriverControlData.h
```diff
@@ -57,14 +46,26 @@ struct MusicInputs
...
+{
+    unsigned char driverControlsBoardAlive;
+    struct LightsInputs lightsInputs;
+    struct MusicInputs musicInputs;
+    unsigned short int acceleration;
+    unsigned short int regenBraking;
+    struct DriverInputs driverInputs;
+};
+
+extern struct DriverControlData driverControlData;
```
bill-luu: This line doesn't have a new line after it, I'm not sure if that matters much, but the other files do.

EpsilonCCS/Src/Data/BatteryData.c
```diff
@@ -0,0 +1,3 @@
+#include "BatteryData.h"
+
+struct BatteryData batteryData = {};
```
JordanHeinrichs: Not sure if this matters much but this initializer form appears to be a gnu extension supported in only a few compilers and it is not standard C. It is however standard C++. Refer to [this](http://en.cppreference.com/w/c/language/struct_initialization) or [this](http://stackoverflow.com/questions/11152160/initializing-a-struct-to-0).  Might be better to change it to `{0}`

PR #15, opened 2017-03-13
fix brakes macro being redefined in driver controls and lights

enochtsang:
It was defined in both DriverControls.h and Lights.h


JordanHeinrichs:
Went ahead with the merge as it looks good, all checks passed, and it is a rather simple change.

PR #18, opened 2017-03-18
Enochtsang/can parsers

enochtsang:


EpsilonCCS/Src/CanParsers/MotorCanParser.c
```diff
@@ -0,0 +1,252 @@
...
+    (*errorData).desaturationFault =
+        data[0] & DESATURATION_FAULT_MASK;
+
+    *rxErrorData = data[7];
+    *txErrorData = data[6];
+}
+
+float arrayToFloat(uint8_t* data)
+{
+    float result =
```
JordanHeinrichs: I this won't work as you are just assigning the integer representation to the float. I.E if you have a floating point number of 16.0, in hex that is 0x41800000. Now that x86 is little endian the internal contents of the float need to end up as 0x00008041 and not just equal to 32833. I would suggest going the union approach from the mbed version of CCS as that is a pretty standard way of converting and much cleaner. If you are opposed to using unions/don't want to use any of the old CCS code then here is another way:

```C
#include <stdio.h>

int main() 
{
   // https://www.h-schmidt.net/FloatConverter/IEEE754.html
   unsigned char data[4] = {0x41, 0x80, 0x00, 0x00}; // 16.0
   float result = 0;
   float resultEnoch = (data[0] << 0) |
      (data[1] << 8) |
      (data[2] << 16) |
      (data[3] << 24);
   
   ((unsigned char*)&result)[0] = data[3];
   ((unsigned char*)&result)[1] = data[2];
   ((unsigned char*)&result)[2] = data[1];
   ((unsigned char*)&result)[3] = data[0];

   printf("result is %f, Enoch method %f\n", result, resultEnoch);
   return 0;
}
```

This outputs
```
jorda00@DESTINY:/mnt/e/Users/jordan/workspace/solarcarTest$ ./a.out
result is 16.000000, Enoch method 32833.000000
```

EpsilonCCS/Src/CanParsers/MpptCanParser.c
```diff
@@ -0,0 +1,35 @@
+#include "cmsis_os.h"
+
+#include "MpptCanParser.h"
+
+#include "MpptData.h"
+
+extern struct MpptData mpptData[NUM_OF_MPPTS];
```
JordanHeinrichs: mpptData is already defined within MpptData.h. It is rather strange that it needs to be externed here as well. Try removing it because I think it might compile without it. This also goes for other the externs within other *Parser.c files.

EpsilonCCS/Src/CanParsers/MotorCanParser.c
```diff
@@ -0,0 +1,252 @@
...
+#define M1_15V_VOLTAGE_RAIL_CAN_ID (0x428)
+#define M1_3_3V_1_9V_VOLTAGE_RAIL_CAN_ID (0x429)
+#define M1_HEATSINK_TEMPERATURE_CAN_ID (0x42B)
+#define M1_DSP_BOARD_TEMPERATURE_CAN_ID (0x42C)
+#define M1_ODOMETER_BUS_AMPHOURS_CAN_ID (0x42E)
+#define M1_SLIP_SPEED_CAN_ID (0x437)
+
+void parseMotorCanMessage(uint32_t stdId, uint8_t* data)
+{
+    switch (stdId)
```
JordanHeinrichs: Optionally you might be able to save quite a bit of duplication by creating a `parseMotorCanMessage(uint32_t messageId, MotorDetailsData* motorDetailsData , uint8_t* data)` that is called from this function. Within this function determine which motor it is from and within the new function make the message comparisons independent of base (maybe pass in stdId - motor#Base and within the new function only use offset from base like the datasheet uses).

EpsilonCCS/Inc/Data/BatteryData.h
```diff
@@ -23,7 +25,7 @@ enum BatteryPrechargeState
 
 struct BatteryData
```
enochtsang: So the goal was to have the CanParsers match the incoming CAN messages, and everything beyond that match the serial protocol. I think its' too much overhead now to change the serial protocol and I don't think there's much benefit anyways.

JordanHeinrichs: :+1: okay sounds good. I didn't realize that the structures matched the outgoing messages.

EpsilonCCS/Src/CanParsers/MotorCanParser.c
enochtsang: Not that I'm opposed to old CCS code, but I didn't like having to do define an additional data structure. Also I can just copy paste from here so I'll do that :P

EpsilonCCS/Src/CanParsers/MotorCanParser.c
enochtsang: Since the IDs are uniquely different, between motor one and motor zero. I think I'll keep this duplication because of the clarity aspect. I would need to use masks and IDs for it to work, I just think the trade off here is worth it. Especially since I really don't expect there to be more motors coming in ever, unlike the amount of MPPTs which is likely to change.

EpsilonCCS/Src/CanParsers/CanParser.c
```diff
@@ -8,43 +8,43 @@
 #include "MpptCanParser.h"
 
 // 0x720 >= AUX_BMS >= 0x723
-#define AUX_BMS_CAN_ID (0x720)
-#define AUX_BMS_CAN_MASK (0xFFC)
+#define AUX_BMS_CAN_MIN (0x720)
+#define AUX_BMS_CAN_MAX (0x72F)
```
JordanHeinrichs: Should the max not be 0x723 as that is the highest CAN message that would be handled for AUX BMS (same idea for the rest)? Or are you going for the full range of CAN addresses that are assigned to AUX BMS? Either way is fine but it is just strange that the comment with the max is different then the maxes that are defined.

enochtsang: Yea I did that to try future proof it more. I removed the comments to fix this.

JordanHeinrichs: Looks good 👍

PR #19, opened 2017-03-21
add task for parsing can messages

enochtsang:
I'll be helping out the UAV team for a bit. so let's merge this one into the CanParsers branch, then merge the CanParsers branch.

### DIFF
--- a/EpsilonCCS/Drivers/STM32F4xx_HAL_Driver/Inc/stm32f4xx_hal_can.h
+++ b/EpsilonCCS/Drivers/STM32F4xx_HAL_Driver/Inc/stm32f4xx_hal_can.h
@@ -409,9 +409,11 @@ typedef struct
 #define CAN_FLAG_TME2              ((uint32_t)0x0000051CU)  /*!< Transmit mailbox 0 empty flag */
 
 /* Receive Flags */
+#define CAN_FLAG_FMP0              ((uint32_t)0x12000003)   /*!< FIFO 0 Message Pending Flag */
 #define CAN_FLAG_FF0               ((uint32_t)0x00000203U)  /*!< FIFO 0 Full flag    */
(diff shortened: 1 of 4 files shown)


enochtsang:
@JordanHeinrichs  For your first point, I'm not really sure. Because what you say about the PoolAlloc and osMessage possibly being slower is true, combining with everything else you said, at this point in time using the queue for the interrupt is not really necessary.

But on the other hand, I've already put in the work for this and I think organizationally I like it better because the interrupt code is very confined now. Say there are issues later with the interrupt execution being too long, if the execution code is small it's easier to debug, compared to having to check across 8+ files if I stick the call the CanParser.c in there.

So I'm undecided which direction to take, why don't you suggest one now and I'll take it haha.

For comment about the driver, I know what I've written works because I've used it in the LightsBoard and manually verified it.

JordanHeinrichs:
My main concern with this way of doing it is it is a more complex then calling it directly. The code path is less clear as your incoming data is being put with the queue.  That said there are benefits of the way you did it. If for an example you wish to put an IO with the can handler for debugging you are now able to. As this is already implemented I think it would be fine to keep it as is. With more complexity there are drawbacks but the changes are valid so feel free to merge this in.On Mar 22, 2017 13:04, Enoch Tsang <notifications@github.com> wrote:@JordanHeinrichs  For your first point, I'm not really sure. Because what you say about the PoolAlloc and osMessage possibly being slower is true, combining with everything else you said, at this point in time using the queue for the interrupt is not really necessary.
But on the other hand, I've already put in the work for this and I think organizationally I like it better because the interrupt code is very confined now. Say there are issues later with the interrupt execution being too long, if the execution code is small it's easier to debug, compared to having to check across 8+ files if I stick the call the CanParser.c in there.
So I'm undecided which direction to take, why don't you suggest one now and I'll take it haha.
For comment about the driver, I know what I've written works because I've used it in the LightsBoard and manually verified it.

PR #21, opened 2017-03-22
Enochtsang/telemetry reporting stub

enochtsang:
After this is merged I can start working on the actual sending part in smaller pieces.

### DIFF
--- a/EpsilonCCS/Inc/TelemetryReporting/TelemetryUtils.h
+++ b/EpsilonCCS/Inc/TelemetryReporting/TelemetryUtils.h
@@ -4,14 +4,23 @@
 #define FRAMING_LENGTH_INCREASE (2)
 #define TERMINATING_BYTE (0x00)
 
+union DataUnion
+{
+    float floatData;
+    short shortData[2];
+    unsigned short uShortData[2];
+    char charData[4];
+    unsigned int uIntData;
+};
+
 unsigned short calculateCrc16(unsigned char* input, int length);
 
-unsigned int frameData(const unsigned char* dataToEncode, unsigned long length, unsigned char* frameData);
-unsigned int stuffData(const unsigned char* dataToEncode, unsigned long length, unsigned char* encodedData);
+unsigned int frameData(unsigned char* dataToEncode, unsigned long length, unsigned char* frameData);
+unsigned int stuffData(unsigned char* dataToEncode, unsigned long length, unsigned char* encodedData);
 void addChecksum(unsigned char* data, unsigned int length);
 
-void writeFloatIntoArray(unsigned char* data, int index, const float& value);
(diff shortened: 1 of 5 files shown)


EpsilonCCS/Src/main.c
```diff
@@ -140,10 +142,12 @@ int main(void)
     /* USER CODE END RTOS_TIMERS */
 
     /* Create the thread(s) */
-    /* definition and creation of defaultTask */
     osThreadDef(canRxTask, parseCanTask, osPriorityNormal, 1, configMINIMAL_STACK_SIZE);
     parseCanHandle = osThreadCreate(osThread(canRxTask), NULL);
 
+    osThreadDef(telemetryOutTask, sendTelemetryTask, osPriorityNormal, 1, configMINIMAL_STACK_SIZE);
```
JordanHeinrichs: This is probably fine as is currently but I think it is rather likely that the stack size will need to be increased as more code is added to this.

EpsilonCCS/Inc/TelemetryReporting/TelemetryUtils.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#define CHECKSUM_LENGTH (2)
+#define FRAMING_LENGTH_INCREASE (2)
+#define TERMINATING_BYTE (0x00)
+
+union DataUnion
```
JordanHeinrichs: This union could be moved to the .c file, other files don't need to reference it. Although it is fine here as well if you wish to keep all types defined within a header file.

EpsilonCCS/Src/main.c
enochtsang: How do you even determine how much stack size you need. Just keep adding it until it starts working?

EpsilonCCS/Inc/TelemetryReporting/TelemetryUtils.h
enochtsang: You know I thought about that, and put it in the header file because it's more "proper". But you're right, it's not going to be used by anything else, so better to not muddy up the namespace.

JordanHeinrichs:
There is probably a better way but that in how I do it.On Mar 23, 2017 00:23, Enoch Tsang <notifications@github.com> wrote:@enochtsang commented on this pull request.



In EpsilonCCS/Src/main.c:
>      osThreadDef(canRxTask, parseCanTask, osPriorityNormal, 1, configMINIMAL_STACK_SIZE);
     parseCanHandle = osThreadCreate(osThread(canRxTask), NULL);
 
+    osThreadDef(telemetryOutTask, sendTelemetryTask, osPriorityNormal, 1, configMINIMAL_STACK_SIZE);

How do you even determine how much stack size you need. Just keep adding it until it starts working?

PR #22, opened 2017-03-24
Send All

enochtsang:
added main sending code, no details on how it's sending yet

### DIFF
--- a/EpsilonCCS/Src/TelemetryReporting/TelemetryReporting.c
+++ b/EpsilonCCS/Src/TelemetryReporting/TelemetryReporting.c
@@ -1,9 +1,48 @@
+#include "cmsis_os.h"
+#include "stm32f4xx.h"
+
 #include "TelemetryReporting.h"
+
+#define CCS_TELEM_FREQ 200 // 5Hz == 200ms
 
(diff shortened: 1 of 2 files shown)


EpsilonCCS/Src/TelemetryReporting/TelemetryReporting.c
```diff
@@ -1,9 +1,48 @@
...
+
 void sendTelemetryTask()
 {
+    uint32_t prevWakeTime = osKernelSysTick();
+    // Savlos defined in https://tinyurl.com/n23nzes
+    static unsigned char salvo = 1;
+
     for (;;)
     {
+        osDelayUntil(&prevWakeTime, CCS_TELEM_FREQ);
```
JordanHeinrichs: This is just something to think about but right now all the data is basically sent in one burst, that might cause issues with the data rate of the radio/uart connection (send buffer might not have enough space). A way around this is to increase the frequency of the thread but not send all every time.

enochtsang: I'll fix that in a later PR.

PR #24, opened 2017-03-24
Interface for transmitting serial

enochtsang:
Who knows if this works ¯\_(ツ)_/¯

I followed this example https://github.com/fboris/STM32Cube_FW_F4/tree/master/Projects/STM32F4-Discovery/Examples/UART/UART_TwoBoards_ComPolling


EpsilonCCS/Src/TransmitSerial.c
```diff
@@ -0,0 +1,23 @@
+#include "TransmitSerial.h"
+
+#include "stm32f4xx_hal_uart.h"
+
+static unsigned char UartReady = 1;
+
+void transmitMessage(uint8_t* payload, uint16_t bufferSize)
```
JordanHeinrichs: This function/class is making a non-blocking DMA transmit into a blocking one, would it instead be possible to use HAL_UART_Transmit() which is a blocking transmit function? If you did that then you wouldn't have to use the UartReady flag and callback. This is just burning CPU cycles/energy while it is waiting.

PR #28, opened 2017-07-24
Bill luu/send faults

bill-luu:


EpsilonCCS/Src/TelemetryReporting/TelemetryReporting.c
```diff
@@ -164,12 +169,35 @@ void sendDriverControls()
...
 }
 
 void sendBatteryFaults()
 {
+    unsigned int unframedPacketLength = BATTERY_FAULTS_LENGTH + CHECKSUM_LENGTH;
+    unsigned char packetPayload[unframedPacketLength];
 
+    packetPayload[0] = BATTERY_FAULTS_PKG_ID;
+    writeBoolsIntoArray(packetPayload, 1, &batteryFaultsData.batteryErrorFlags, 21);
+    writeBoolsIntoArray(packetPayload, 2, &batteryFaultsData.batteryLimitFlags, 14);
```
JordanHeinrichs: I think the index should be 5 as per the docs, the flags above is a uint (4 bytes),

EpsilonCCS/Src/TelemetryReporting/TelemetryReporting.c
```diff
@@ -7,10 +7,15 @@
...
 #include "MotorDetailsData.h"
 #include "DriverControlData.h"
+#include "MotorFaultsData.h"
+#include "BatteryFaultsData.h"
 
 #define KEY_MOTOR_LENGTH (43)
 #define MOTOR_DETAILS_LENGTH (69)
 #define DRIVER_CONTROLS_LENGTH (9)
+#define MOTOR_FAULTS_LENGTH (9)
+#define BATTERY_FAULTS_LENGTH (3)
```
JordanHeinrichs: Should be 7 I think

bill-luu: I wasn't really sure about this, since I was basing it off of `KEY_MOTOR_LENGTH` being 4 bytes shorter than on the protocol and figured there was 4 bytes of metadata?

JordanHeinrichs: KEY_MOTOR_LENGTH is wrong in the protocol, it should just be 43 bytes long.

Battery faults length is correct within the protocol. This reasoning is 1 byte for package ID, 4 bytes for battery faults (21 faults which mean 21 bits. With 8 bits in a byte that is at minimum 3 bytes. For alignment and general ease that is just rounded up to 4 bytes). 2 Bytes for the Limit flags (16 flags, 16 bits, 2 bytes)

bill-luu: I think KEY_MOTOR_LENGTH in here is correct, but the protocol sizes (43)is including the Overhead Byte(1) + Checksum(2) + Frame Delimiter(1) = 47 Bytes.

So yeah you're right, BATTERY_FAULTS_LENGTH should be 7 here, but on the protocol it should be 11 (Just edited it).

PR #30, opened 2017-07-29
Send Mppt

bill-luu:
WIP
-Still need to do aliveBit (Do I need to write the alive bit onto bit 7 of each mpptStatus byte? Or is that just already there)
-Should Mppt Size on the protocol be 14 instead of 13? (Last index 8 + short unit 2 + metadata 4)


enochtsang:
Isn't CI a glorious thing

bill-luu:
Descriptions are pretty sick too

enochtsang:
Uh no, 13 seems right. Last index 8 + 1 for short uint. index 8 and 9 is the last set of info. So size of data is 9 bytes + 4 meta data. 13

bill-luu:
Wait, but if Motor Faults ended at index 8 + uchar(1 byte) + metadata = 13, how does MPPT have index 8 + short uint(2 bytes) + metadata = 13?

bill-luu:
Also, what about the alive bit

enochtsang:
Nvm yea you're right should be 14 for MPPT.

bill-luu:
What about the alive bit

bill-luu:
Ready to review!

PR #31, opened 2017-08-16
Bill luu/send lights

bill-luu:


EpsilonCCS/Src/TelemetryReporting/TelemetryReporting.c
```diff
@@ -19,6 +20,7 @@
 #define BATTERY_FAULTS_LENGTH (3)
 #define BATTERY_DETAILS_LENGTH (51)
 #define MPPT_DETAILS_LENGTH (13)
+#define LIGHTS_DETAILS_LENGTH (7)
```
JordanHeinrichs: The message has only 3 bytes of content. While this matches the protocol I think the protocol is wrong here.

bill-luu: I think having the 3 uchars makes sense, unless we're missing anything in the protocol @enochtsang ?

enochtsang: Yea, 3 bytes content but +4 overhead

JordanHeinrichs:
The merge conflicts resolution looks good, the lights packet length still should be looked at. It doesn't seem right for it to be 7 bytes long when there is only 3 bytes of content.

bill-luu:
Yeah you're right, will fix tomorrow.

PR #33, opened 2017-09-06
Added send rtr packet

bill-luu:
[Link to Mppt Datasheet](http://calgarysolarcarteamhub.com/attachments/download/112/Photon%20Datasheet%20rev1_2.pdf)

So for this, I'm not entirely sure if I'm doing this right at all. I'm mainly concerned on how they want the Rtr package to be structured. I think it's just saying 0x600 + the binary encoded channel of the mppt? I'm not sure if I need anything else.

Also, is it better to send the packets of each channel all in one function, or should I send it in 3 different functions?


JordanHeinrichs:
Took a glance over it and it seems good. I will take another look tomorrow. Been staring at code for over 12 hours today and everything is a blur now.

EpsilonCCS/Inc/MpptRtr.h
```diff
@@ -0,0 +1,20 @@
...
+
+#define MPPT_HEARTBEAT_CAN_FREQ 200
+#define MPPT_STDID 0x600U
+#define MPPT_DLC 8
+
+extern CAN_HandleTypeDef hcan1;
+extern osPoolId canTxPool;
+extern osMessageQId canTxQueue;
+
+typedef struct // Datasheet is here: http://calgarysolarcarteamhub.com/attachments/download/112/Photon%20Datasheet%20rev1_2.pdf
```
enochtsang: Can you move this to the drive.

EpsilonCCS/Inc/MpptRtr.h
```diff
@@ -0,0 +1,20 @@
...
+{
+    uint32_t StdId;
+    uint32_t Channel; //not sure of the data type, but it is a binary encoded number from 0-3.
+} MpptCanMsg;
+
+void mpptRtrTask(void const* arg);
+void sendCanTask(void const* arg);
```
enochtsang: Missing newline.

EpsilonCCS/Inc/MpptRtr.h
```diff
@@ -0,0 +1,20 @@
...
+#define MPPT_DLC 8
+
+extern CAN_HandleTypeDef hcan1;
+extern osPoolId canTxPool;
+extern osMessageQId canTxQueue;
+
+typedef struct // Datasheet is here: http://calgarysolarcarteamhub.com/attachments/download/112/Photon%20Datasheet%20rev1_2.pdf
+{
+    uint32_t StdId;
+    uint32_t Channel; //not sure of the data type, but it is a binary encoded number from 0-3.
```
enochtsang: These two are member variables, should be capCase.

EpsilonCCS/Src/MpptRtr.c
```diff
@@ -0,0 +1,44 @@
...
+    for (;;)
+    {
+        osDelayUntil(&prevWakeTime, MPPT_HEARTBEAT_CAN_FREQ);
+        // Allocate CAN Message, deallocated by sender "sendCanTask()"
+        MpptCanMsg* msg = (MpptCanMsg*)osPoolAlloc(canTxPool);
+        // Populate CAN Message
+        msg->StdId = MPPT_STDID;
+        msg->Channel = channel;
+        // Update Channel
+        channel = (channel++) % 3;
```
enochtsang: Can you set it back to 0 after 3 instead, prevent overflow.

EpsilonCCS/Src/MpptRtr.c
```diff
@@ -0,0 +1,44 @@
...
+        // Send CAN Message
+        osMessagePut(canTxQueue, (uint32_t)msg, osWaitForever);
+    }
+}
+
+void sendCanTask(void const* arg)
```
enochtsang: This function doesn't look like it's being used anywhere.

EpsilonCCS/Src/MpptRtr.c
bill-luu: Done. Although wouldn't modulus do that anyways? When channel = 3, channel = 3 % 3 = 0?

enochtsang: Ah ya you're right, I misread this. Would've been fine as is.

EpsilonCCS/Inc/MpptRtr.h
```diff
@@ -0,0 +1,20 @@
...
+extern osMessageQId canTxQueue;
+
+typedef struct // Datasheet is here: https://drive.google.com/open?id=0B4Qf5FVsKl-SVDRFY3RoWUt0RVU
+{
+    uint32_t stdId;
+    uint32_t channel; //not sure of the data type, but it is a binary encoded number from 0-3.
+} MpptCanMsg;
+
+void mpptRtrTask(void const* arg);
+void sendCanTask(void const* arg);
```
enochtsang: Sending CAN is not directly coupled with the MPPT logic, we can also potentially need to send CAN messages that are not MPPT related, so it'd be better to have this in a separate file especially because the naming and usage of this function is so generic.

EpsilonCCS/Src/MpptRtr.c
```diff
@@ -0,0 +1,53 @@
...
+            hcan1.pTxMsg->StdId = msg->stdId;
+            hcan1.pTxMsg->DLC = MPPT_DLC;
+            // Send CAN Message
+            HAL_GPIO_TogglePin(LED_RED_GPIO_Port, LED_RED_Pin);
+            HAL_CAN_Transmit_IT(&hcan1);
+            // Deallocate CAN message
+            osPoolFree(canTxPool, msg);
+        }
+    }
+}
```
enochtsang: Need a new line.

EpsilonCCS/Src/MpptRtr.c
```diff
@@ -0,0 +1,53 @@
...
+    {
+        osDelayUntil(&prevWakeTime, MPPT_HEARTBEAT_CAN_FREQ);
+        // Allocate CAN Message, deallocated by sender "sendCanTask()"
+        MpptCanMsg* msg = (MpptCanMsg*)osPoolAlloc(canTxPool);
+        // Populate CAN Message
+        msg->stdId = MPPT_STDID;
+        msg->channel = channel;
+
+        // Update Channel
+        if (channel = 3)
```
enochtsang: Sorry man, but change it back it looked better before haha (I was wrong thinking it would overflow before).

EpsilonCCS/Inc/MpptRtr.h
bill-luu: Should I keep the mppt can message a separate structure from the regular CAN messages? I would say that it has a more unique thing with the channel value.

bill-luu:
So it turns out that each channel is actually on it's own address, i.e Channel Zero is `0x600`, One is `0x601`, Two is `0x602`. We'll definitely need to test this but for now I think it's good to review.

I updated the communication protocol accordingly.

PR #38, opened 2017-11-25
Lights Board - Headlights & Signal Fixes

shilpan97:
Needed to make changes to headlights high and headlights low and when they are both turned on. This is response to testing the headlights using CANAnalyzer. 
Checkout the Google Doc for testing results when initially done. 
https://docs.google.com/spreadsheets/d/1LpC1YUID1vZMPWIQ9MZZMY8twGFfwgsB3J8d-bpGb4g/edit#gid=0 

Forgot to remove some additional white spaces..will fix that.

### DIFF
--- a/EpsilonLights/Inc/Lights.h
+++ b/EpsilonLights/Inc/Lights.h
@@ -31,7 +31,7 @@
 #define BRAKES_INPUT_INDEX_P1 3 // 24 = [*3*] >> 0
 #define BRAKES_INPUT_INDEX_P2 0 // 24 = [3] >> *0*
 
-#define BATTERY_STAT_STDID 0x303
+#define BATTERY_STAT_ERRORS_STDID 0x303
 // Mask for to check faults defined for Tritium BMS in TRI67.010 ver 2
 // Check for "cell over voltage" 0x01, "cell under voltage" 0x02, and "cell over temperature" 0x04
 #define BATTERY_CRIT_FAULT_MASK 0x07
@@ -54,9 +54,8 @@ typedef struct SigLightsHandle
 
 extern CAN_HandleTypeDef hcan2; // main.c
 extern uint8_t lightsInputs;
-extern uint8_t batteryErrors;
+extern uint8_t batteryErrors[5];
 extern uint8_t driversInputs[4];
-extern uint8_t batteryStatus[4];
 extern SigLightsHandle sigLightsHandle;
 
 // Task for updating GPIOs
--- a/EpsilonLights/Inc/Lights.h
+++ b/EpsilonLights/Inc/Lights.h
@@ -31,7 +31,7 @@
 #define BRAKES_INPUT_INDEX_P1 3 // 24 = [*3*] >> 0
 #define BRAKES_INPUT_INDEX_P2 0 // 24 = [3] >> *0*
 
-#define BATTERY_STAT_ERRORS_STDID 0x303
+#define BATTERY_STAT_ERRORS_STDID 0x303U
 // Mask for to check faults defined for Tritium BMS in TRI67.010 ver 2
 // Check for "cell over voltage" 0x01, "cell under voltage" 0x02, and "cell over temperature" 0x04
 #define BATTERY_CRIT_FAULT_MASK 0x07
--- a/EpsilonLights/Src/Lights.c
+++ b/EpsilonLights/Src/Lights.c
@@ -20,15 +20,15 @@ void updateLightsTask(void const* arg)
     {
         osDelayUntil(&prevWakeTime, LIGHTS_UPDATE_FREQ);
         headlightsOff = (lightsInputs >> HOFF_INPUT_INDEX) & 1;
-        headlightsLow = (lightsInputs >> HLOW_INPUT_INDEX) & 1;
-        headlightsHigh = (lightsInputs >> HHIGH_INPUT_INDEX) & 1;
+        headlightsLow = (lightsInputs >> HLOW_INPUT_INDEX) & 1; //0b0000x0010
+        headlightsHigh = (lightsInputs >> HHIGH_INPUT_INDEX) & 1; //0b0000x0100
         rightSignal = (lightsInputs >> RSIGNAL_INPUT_INDEX) & 1;
         leftSignal = (lightsInputs >> LSIGNAL_INPUT_INDEX) & 1;
         hazards = (lightsInputs >> HAZARDS_INPUT_INDEX) & 1;
 
         /* UPDATE HEADLIGHTS */
 
-        if (headlightsOff && headlightsLow + headlightsHigh)
+        if ((headlightsOff) && (headlightsLow + headlightsHigh))
         {
             // Error state, turn headlights off
             HAL_GPIO_WritePin(HHIGH_GPIO_Port, HHIGH_Pin, LIGHT_OFF);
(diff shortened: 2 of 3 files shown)


bill-luu:
Run this through astyle

EpsilonLights/Src/Lights.c
```diff
@@ -40,6 +40,38 @@ void updateLightsTask(void const* arg)
...
+        {
+            HAL_GPIO_WritePin(HHIGH_GPIO_Port, HHIGH_Pin, LIGHT_ON);
+        }
+        else
+        {
+            HAL_GPIO_WritePin(HHIGH_GPIO_Port, HHIGH_Pin, LIGHT_OFF);
+        }
+
+
+
```
bill-luu: Lots of extra whitespace here.

EpsilonLights/Src/Lights.c
```diff
@@ -20,15 +20,15 @@ void updateLightsTask(void const* arg)
     {
         osDelayUntil(&prevWakeTime, LIGHTS_UPDATE_FREQ);
         headlightsOff = (lightsInputs >> HOFF_INPUT_INDEX) & 1;
-        headlightsLow = (lightsInputs >> HLOW_INPUT_INDEX) & 1;
-        headlightsHigh = (lightsInputs >> HHIGH_INPUT_INDEX) & 1;
+        headlightsLow = (lightsInputs >> HLOW_INPUT_INDEX) & 1; //0b0000x0010
+        headlightsHigh = (lightsInputs >> HHIGH_INPUT_INDEX) & 1; //0b0000x0100
```
bill-luu: Remove these comments

enochtsang: Orr they can be expanded on to be useful :)

shilpan97:
Fixed on Bill's computer.. the headlightlow and headlightshigh case had to be moved to the beginning of the code.. retested on Nov 28 and all cases work now.

enochtsang:
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

EpsilonLights/Src/Lights.c
```diff
@@ -20,15 +20,44 @@ void updateLightsTask(void const* arg)
...
+        if (headlightsHigh)
+        {
+            HAL_GPIO_WritePin(HHIGH_GPIO_Port, HHIGH_Pin, LIGHT_ON);
+        }
+        else
+        {
+            HAL_GPIO_WritePin(HHIGH_GPIO_Port, HHIGH_Pin, LIGHT_OFF);
+        }
+
+        if ((headlightsOff) && (headlightsLow || headlightsHigh))
```
bill-luu: Imo I think it would make more sense to do 
````
if(headlightsOff)
{
 Turn off lights
}
else if (headlightsHigh && headlightsLow)
{
 Turn off headlights
}
else()
{
 Turn on/off headlights as needed
}
````

That way we don't write the headlights first before checking if headlights are off.

File: EpsilonLights/Src/Lights.c [hunk 4]
```diff
@@ -40,6 +69,10 @@ void updateLightsTask(void const* arg)
             HAL_GPIO_WritePin(HLOW_GPIO_Port, HLOW_Pin, !headlightsLow);
         }
 
+
+
+
+
```
bill-luu: Still lots of whitespace here

shilpan97:
Changed the CAN address/identifier since it was not working for the other cases(aka batteryErrors).  
Bill and I believe it may have something to do with the FiterIdHigh and FiterIdLow for filtering the registers.  It seems to be working when we swap the addresses for LightsInputs with batteryErrors. LightsInputs is set to another CAN identifier which works.

bill-luu:
All tests now pass except for the BMS Strobe light. I'll address that in a later PR, please review when you can!
Merry Christmas!

EpsilonLights/Src/Lights.c
```diff
@@ -27,8 +27,13 @@ void updateLightsTask(void const* arg)
...
         /* UPDATE HEADLIGHTS */
-
-        if (headlightsOff && headlightsLow + headlightsHigh)
+        if ((headlightsOff) && (headlightsLow || headlightsHigh))
+        {
+            // Error state, turn headlights off
+            HAL_GPIO_WritePin(HHIGH_GPIO_Port, HHIGH_Pin, LIGHT_OFF);
+            HAL_GPIO_WritePin(HLOW_GPIO_Port, HLOW_Pin, LIGHT_OFF);
+        }
+        else if (headlightsLow && headlightsHigh)
```
JordanHeinrichs: For safety, if it gets into an error state, wouldn't it be better to leave the headlights on? If it is daylight and the headlights turn on it wastes some power but not that bad. If it is dark out and it enters error state it would be dangerous to turn off all lights.

EpsilonLights/Inc/Lights.h
```diff
@@ -31,7 +31,7 @@
 #define BRAKES_INPUT_INDEX_P1 3 // 24 = [*3*] >> 0
 #define BRAKES_INPUT_INDEX_P2 0 // 24 = [3] >> *0*
 
-#define BATTERY_STAT_STDID 0x6FBU
+#define BATTERY_STAT_ERRORS_STDID 0x303
```
enochtsang: Can you explain this change?

EpsilonLights/Src/Lights.c
enochtsang: I think this checking for an error state could be written better. Ultimately what's being checked for is that there is no more than 1 of `headlightsOff`, `headlightsLow` and `headlightsHigh` is on at any given point in time. How's it's written now doesn't feel like it's directly checking that.

EpsilonLights/Src/Lights.c
```diff
@@ -225,23 +239,28 @@ void HAL_CAN_RxCpltCallback(CAN_HandleTypeDef* hcan)
...
+        // Memory is stored in Little Endian format
+        batteryErrors[0] = msg->Data[4];
+        batteryErrors[1] = msg->Data[3];
+        batteryErrors[2] = msg->Data[2];
+        batteryErrors[3] = msg->Data[1];
+        batteryErrors[4] = msg->Data[0];
     }
-    else if (msg->StdId == BATTERY_STAT_STDID && msg->DLC == 4)
+
+    else if (msg->StdId == DRIVERS_INPUTS_STDID && msg->DLC == 4)
```
enochtsang: The spaces above the else ifs are misleading, please remove them.

EpsilonLights/Src/Lights.c
```diff
@@ -225,23 +239,28 @@ void HAL_CAN_RxCpltCallback(CAN_HandleTypeDef* hcan)
     CanRxMsgTypeDef* msg = hcan->pRxMsg;
     HAL_GPIO_TogglePin(LED_RED_GPIO_Port, LED_RED_Pin);
 
-    if (msg->StdId == LIGHTS_INPUT_STDID)
+    if (msg->StdId == LIGHTS_INPUT_STDID && msg->DLC == 1)
     {
         lightsInputs = msg->Data[0];
     }
-    else if (msg->StdId == DRIVERS_INPUTS_STDID)
```
enochtsang: In the future, don't mess with the order of the else ifs if you don't need to, this has caused an unnecessarily complicated diff, when this diff in `HAL_CAN_RxCpltCallback()` could've been a lot less verbose.

EpsilonLights/Src/main.c
```diff
@@ -44,8 +44,8 @@ CAN_HandleTypeDef hcan2;
 /* USER CODE BEGIN PV */
 /* Private variables ---------------------------------------------------------*/
 uint8_t lightsInputs; // Initialized to 0
+uint8_t batteryErrors[5]; //Initialized to {0,0,0,0}
```
enochtsang: This comment is not correct

EpsilonLights/Src/main.c
```diff
@@ -287,12 +287,29 @@ static void MX_CAN2_UserInit(void)
     sFilterConfig.FilterActivation = ENABLE;
     sFilterConfig.BankNumber = 0; // Set all filter banks for CAN2
 
+    CAN_FilterConfTypeDef batteryFilterConfig;
+    batteryFilterConfig.FilterNumber = 1; // Use secondary filter bank
+    batteryFilterConfig.FilterMode = CAN_FILTERMODE_IDLIST; // Look for specific can messages
+    batteryFilterConfig.FilterIdHigh = BATTERY_STAT_ERRORS_STDID << 5; // Filter registers need to be shifted left 5 bits
+    batteryFilterConfig.FilterIdLow = 0; // Filter registers need to be shifted left 5 bits
+    batteryFilterConfig.FilterMaskIdHigh = DRIVERS_INPUTS_STDID << 5; // Unused
```
enochtsang: This comment is incorrect

EpsilonLights/Src/main.c
```diff
@@ -287,12 +287,29 @@ static void MX_CAN2_UserInit(void)
     sFilterConfig.FilterActivation = ENABLE;
     sFilterConfig.BankNumber = 0; // Set all filter banks for CAN2
 
+    CAN_FilterConfTypeDef batteryFilterConfig;
```
enochtsang: Why is a second filter necessary?

EpsilonLights/Src/Lights.c
```diff
@@ -65,15 +76,18 @@ void updateLightsTask(void const* arg)
...
-            HAL_GPIO_WritePin(ESTROBE_GPIO_Port, ESTROBE_Pin, LIGHT_ON);
-        }
-        else
-        {
-            HAL_GPIO_WritePin(ESTROBE_GPIO_Port, ESTROBE_Pin, LIGHT_OFF);
-        }
+        /*Update BMS Strobe*/
+        HAL_GPIO_WritePin(ESTROBE_GPIO_Port, ESTROBE_Pin, LIGHT_OFF);
+        // TODO Parse the Error Messages and turn on the BMS on certain messages
+        // Strobe the BMS
```
enochtsang: We don't need this commented code

EpsilonLights/Src/main.c
bill-luu: Here's the explanation:
https://github.com/UCSolarCarTeam/Epsilon-Embedded-Software/issues/40

EpsilonLights/Inc/Lights.h
bill-luu: The new Orion BMS has the errors on ID 0x303

bill-luu: I did forget to add in U though

EpsilonLights/Src/Lights.c
bill-luu: Fair enough!

enochtsang:
Looks ok to me, can someone QA this before we merge it?

Issue #34, opened 2017-10-12
AUX signal in driver controls


bill-luu:
So I had a chat with the electrical managers, and they let me know that there is an additional input going into the driver controls called the Aux input (Not to be confused with the Audio Aux). That represents the first state on the control dial on the car (The others are Forward and Reverse). When the Aux signal is sent, that tells the car that the 12V line should be connected, but not the high voltage one. This should be put into CAN for the Aux BMS to handle.

We need to: 

1. Update the Protocol

1. Send this signal via CAN

1. Propagate the changes through to Serial/Json and corresponding programs (Hermes, Test Tool, Dashboard)

bill-luu:
Actually, looking closer at the schematic, there is no music aux! We are accounting for it, however the AUX signal is getting sent as a part of music inputs :/

PR #39, opened 2017-11-29
Moved Aux from Music player to Drivers and changed subsequent CAN data

Reddragoneer1:
Moved Aux from Music player to Driver. Previous Music player had Can data in hexadecimal with 1, 2, 3, 4, 10. 3 should not be there because it would mean 2 outputs, Aux and Volume Up.

### DIFF
--- a/EpsilonCCS/Inc/Data/DriverControlData.h
+++ b/EpsilonCCS/Inc/Data/DriverControlData.h
@@ -56,6 +56,7 @@ struct DriverInputs
     unsigned char pushToTalk;
     unsigned char horn;
     unsigned char reset;
+    unsigned char aux;
 };
 
 struct DriverControlData
@@ -66,7 +67,6 @@ struct DriverControlData
     unsigned short int acceleration;
     unsigned short int regenBraking;
     struct DriverInputs driverInputs;
-    unsigned char aux;
 };
 
 extern struct DriverControlData driverControlData;
--- a/EpsilonCCS/Inc/Data/DriverControlData.h
+++ b/EpsilonCCS/Inc/Data/DriverControlData.h
@@ -15,11 +15,10 @@
 #define INTERIOR_MASK 0x40
 
 // Music Inputs
-#define AUX_MASK 0x01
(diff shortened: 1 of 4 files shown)


EpsilonCCS/Inc/Data/DriverControlData.h
```diff
@@ -15,11 +15,10 @@
...
 // Music Inputs
-#define AUX_MASK 0x01
-#define VOLUME_UP_MASK 0x02
-#define VOLUME_DOWN_MASK 0x04
-#define NEXT_SONG_MASK 0x08
-#define PREV_SONG_MASK 0x10
+#define VOLUME_UP_MASK 0x01
+#define VOLUME_DOWN_MASK 0x02
+#define NEXT_SONG_MASK 0x04
+#define PREV_SONG_MASK 0x8
```
bill-luu: Use `0x08` for consistency with the rest.

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -121,7 +120,7 @@ void sendDriverTask(void const* arg)
         msg->Data[3] |= 0x08 * !HAL_GPIO_ReadPin(PUSH_TO_TALK_GPIO_Port, PUSH_TO_TALK_Pin);
         msg->Data[3] |= 0x10 * !HAL_GPIO_ReadPin(HORN_GPIO_Port, HORN_Pin);
         msg->Data[3] |= 0x20 * !HAL_GPIO_ReadPin(RESET_GPIO_Port, RESET_Pin);
-        // Send CAN Message
```
bill-luu: The comment that was here got removed, let's put it back.

EpsilonCCS/Inc/Data/DriverControlData.h
Reddragoneer1: Oh, oops

bill-luu:
Nice job!

In the future, add "Fixes #<issueNumber>" to the PR description to link it with the issue, so that when the PR is merged, the issue is closed as well!

In your case, this would close #34

PR #45, opened 2018-01-04
switched the left and right signals to match the protocol

Reddragoneer1:
Lines 185 and 186

### DIFF
--- a/EpsilonLights/Src/Lights.c
+++ b/EpsilonLights/Src/Lights.c
@@ -182,8 +182,8 @@ void reportLightsToCanTask(void const* arg)
         hcan2.pTxMsg->Data[0] += !stat.lowBeams * 0x01;
         hcan2.pTxMsg->Data[0] += !stat.highBeams * 0x02;
         hcan2.pTxMsg->Data[0] += !stat.brakes * 0x04;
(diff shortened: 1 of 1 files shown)


bill-luu:
Oh shoot I totally didn't realize this was a thing! Great catch :) 
`Lights.h` needs to be updated as well with the switch.

Does CCS & Driver controls have this issue as well?

bill-luu:
If they all have it, we might as well change the protocol

bill-luu:
Actually, disregard the comment I made about switching the signals in `Lights.h`, that's reading the values of driver inputs correctly. Just this should be OK. 

You managed to reveal that one of our tests for the lights were wrong! Our tests had the signal lights inputs as right = 4, left = 3, when it should have matched the right = 3, left = 4 in the driver controls. If we hadn't caught that, our data on the lights would have been reversed D:.

Hey @enochtsang , tldr our communications protocol has the signals lights reversed on the lights output (left on index 3, right on 4) compared to the driver controls (left 4, right 3). We can leave the protocol and merge this PR and be good (possibly leave future maintainers confused) or change it to be consistent (Update the driver controls & CCS as well). What do you think?

bill-luu:
Oh and last comment for now, don't forget to assign reviewers!

bill-luu:
Continued in #46

PR #51, opened 2018-02-10
Devonvipond/blink bms lights

DevonVipond:


EpsilonLights/Src/main.c
```diff
@@ -289,11 +294,12 @@ static void MX_CAN2_UserInit(void)
 
     CAN_FilterConfTypeDef batteryFilterConfig;
     batteryFilterConfig.FilterNumber = 1; // Use secondary filter bank
+    batteryFilterConfig.FilterScale = CAN_FILTERSCALE_32BIT;
     batteryFilterConfig.FilterMode = CAN_FILTERMODE_IDLIST; // Look for specific can messages
-    batteryFilterConfig.FilterIdHigh = BATTERY_STAT_ERRORS_STDID << 5; // Filter registers need to be shifted left 5 bits
```
enochtsang: Why was this removed? We need to be checkin this for the strobe light, what's changed?

EpsilonLights/Src/main.c
```diff
@@ -126,6 +128,9 @@ int main(void)
     // Setup task to handle blinking left and right signal lights
     osThreadDef(blinkLightsTask, blinkSignalLightsTask, osPriorityNormal, 1, configMINIMAL_STACK_SIZE);
     blinkLightsHandle = osThreadCreate(osThread(blinkLightsTask), NULL);
+    //Setup task to update strobe light
+    osThreadDef(strobeLightTask, updateStrobeLight, osPriorityNormal, 1, configMINIMAL_STACK_SIZE);
```
enochtsang: Have you checked this works? This stack size might not be enough.

EpsilonLights/Src/Lights.c
```diff
@@ -154,6 +151,44 @@ void blinkSignalLightsTask(void const* arg)
...
+        {
+            HAL_GPIO_WritePin(ESTROBE_GPIO_Port, ESTROBE_Pin, LIGHT_ON);
+        }
+        else
+        {
+            HAL_GPIO_WritePin(ESTROBE_GPIO_Port, ESTROBE_Pin, LIGHT_OFF);
+        }
+
+        // Update blinker timer
+        if (blinkerTimer > BLINKER_FREQ * 2)
```
enochtsang: This doesn't need to be continuously updated, having the timer run continuously also prevents the light from coming on right away. If a fault stage is hit, we should see the strobe light up right away, not the next cycle of the blinker timer.

EpsilonLights/Src/main.c
bill-luu: We added a few things to the comms protocol. Electrical explained that if a fault trips, the Orion contractors will disengage, so the Aux BMS will catch that and notify the lights board.

EpsilonLights/Src/Lights.c
```diff
@@ -154,6 +151,49 @@ void blinkSignalLightsTask(void const* arg)
...
+    // If blinkerTimer is greater than (BLINKER_FREQ*2) reset blinkerTimer to 0
+    uint32_t blinkerTimer = 0;
+    char strobeLight;
+
+    for (;;)
+    {
+        osDelayUntil(&prevWakeTime, LIGHTS_UPDATE_FREQ);
+        //  headlightsOff = (lightsInputs >> HOFF_INPUT_INDEX) & 1;
+        //  brakes = (driversInputs[BRAKES_INPUT_INDEX_P1] >> BRAKES_INPUT_INDEX_P2) & 1;
+        strobeLight = (auxBmsInputs[1] >> 0) & 1;
```
bill-luu: is 1 supposed to be the strobe_fault_mask?

EpsilonLights/Src/Lights.c
```diff
@@ -154,6 +151,49 @@ void blinkSignalLightsTask(void const* arg)
...
+
+        /*Update BMS Strobe*/
+        //  if (auxBmsInputs[0] & STROBE_FAULT_MASK && blinkerTimer >= BLINKER_FREQ)
+        if (strobeLight && blinkerTimer <= BLINKER_FREQ)
```
bill-luu: Add in some brackets, it kinda looks like we're comparing strobeLight && blinkerTimer then seeing if it is less than the freq lol

EpsilonLights/Src/Lights.c
```diff
@@ -251,6 +291,12 @@ void HAL_CAN_RxCpltCallback(CAN_HandleTypeDef* hcan)
         driversInputs[2] = msg->Data[1];
         driversInputs[3] = msg->Data[0];
     }
+    else if (msg->StdId == AUXBMS_INPUT_STDID)// not sure if dlc should equal somthing
```
bill-luu: dlc needs to be equal to the length of the 0x721 message

EpsilonLights/Src/Lights.c
DevonVipond: Yea

DevonVipond: Should i use a variable instead?

bill-luu: Yes please!

bill-luu:
Verified working, just fix up that mask 👍

EpsilonLights/Src/Lights.c
```diff
@@ -154,6 +151,46 @@ void blinkSignalLightsTask(void const* arg)
...
+            HAL_GPIO_WritePin(ESTROBE_GPIO_Port, ESTROBE_Pin, LIGHT_OFF);
+        }
+
+        // Update blinker timer
+        if ((blinkerTimer > BLINKER_FREQ * 2) || !strobeLight)
+        {
+            // If blinkerTimer is greater than (BLINKER_FREQ*2) reset blinkerTimer to 0
+            blinkerTimer = 0;
+        }
+        else
```
enochtsang: There should be one more `else if` where if `!strobeLight`, `blinkerTimer` gets set to 0. That way, the strobe light will come on as soon as there is a trip, and doesn't have to wait for the next cycle.

EpsilonLights/Inc/Lights.h
```diff
@@ -36,6 +36,9 @@
 // Check for "cell over voltage" 0x01, "cell under voltage" 0x02, and "cell over temperature" 0x04
 #define BATTERY_CRIT_FAULT_MASK 0x07
```
enochtsang: Is this still being used?

EpsilonLights/Src/Lights.c
DevonVipond: Im not sure i understand. Doesn't blinker timer already get set to zero if !strobeLight and the strobe light comes on immediately because blinkerTimer will start off as 0?

EpsilonLights/Inc/Lights.h
DevonVipond: Its not. I will remove it.

EpsilonLights/Src/Lights.c
enochtsang: Ah yup you're right, I read it wrong.

PR #52, opened 2018-02-10
Magnus/aux bms can

mglyngberg:
Added Strobe 'BMS Light', 'Allow Charge' and 'Contactor Error' CAN parsing in for the aux BMS, and added those values in the batteryData.


EpsilonCCS/Inc/Data/BatteryData.h
```diff
@@ -50,6 +55,9 @@ struct BatteryData
     unsigned short int averageCellVoltage;
     unsigned char prechargeState;
     unsigned char auxVoltage;
+    unsigned char strobeBMSLight;
```
enochtsang: Why is this from the aux BMS? What value does this hold?

PR #54, opened 2018-02-18
Updated CCS Config

bill-luu:
# Changelog

**Note:** I'm making two PRs, one for the new config, and one for any changes to the CCS files that result from it. This branch might not actually make properly, but the other will definitely. The other is a WIP, so feel free to merge this one first. 

Nothing was working on the previous configuration, and I spent too long trying to figure out why so I re-did the configurations with the latest firmware. You should be able to open up the configuration perfectly from there. LEDs are able to flash!

I also used CubeMx's makefile file generator instead of CubeMX2MakeFile, since it wasn't able to make the project properly. The generated makefile requires that you have to add in each of the files manually and point to the location of gcc. I think we should consider removing the Makefile from the gitignore, since we don't have a generator now.

Highlights of the config:

- vTaskDelayUntil is Enabled
- CAN1 -  Master, CAN2 - Slave
- RCC - High Speed Crystal
- Sys - Timebase (TIM2)
- USART3 - Asynchronous 

Screenshots of the config:
image
image


badeniran:
Looks good. Only question I have is are you sure about the USART configs? If you are, then everything looks good

bill-luu:
Actually that's a good point, I did a bit more reading and I think I'm going to change it to:
USART3 - Asynchronous with Hardware Flow Control as CTS only

Definitions:
image
 
I figure to just have CTS, since RTS isn't relevant if the board isn't expecting to receive data. I could also have it disabled, but idk. @enochtsang what are your thoughts?

bill-luu:
Actually, nvm, CCS Pinout specifies CTS & RTS, so I'll just leave it as CTS/RTS configured.

bill-luu:
How do these USART configs look:
image

PR #55, opened 2018-02-22
CCS Full Functions

bill-luu:
Note: This PR is set to merge to `bill-luu/CCSreconfig`, just so the review diff isn't ridiculous. Once this is approved, I'll close this PR and merge it onto master.

# Changelog

- Properly set up the USART, and now the CCS can transmit to Hermes.
- Verified the MPPT CAN stuff
- Use proper filter banks for reception
- Added in LED indications: Blue for sending MPPT messages, Green for reception, Red for being unable to send USART info.

I'm gonna let @DevonVipond test this boy.

After this PR, the CCS will need to be fully tested with Hermes (I verified that messages will be sent properly, now just to catch for small mistakes)


EpsilonCCS/Src/TransmitSerial.c
```diff
@@ -2,7 +2,37 @@
 
 #include "stm32f4xx_hal_uart.h"
 
+/*
+* Flashing Red LED: The USART line is has errored sending this message, will re-try
+* Solid Red LED: The USART Port has timed out, check if Hermes is on
```
enochtsang: Some other program may be listening in the future, it's more future proof to not says Hermes but like "other end of serial port is initialized" or something like that.

EpsilonCCS/Src/TransmitSerial.c
```diff
@@ -2,7 +2,37 @@
...
+*/
 void transmitMessage(uint8_t* payload, uint16_t bufferSize)
 {
-    HAL_UART_Transmit(&huart3, payload, bufferSize, 5000);
+    uint16_t attemptCount = 0;
+
+    while (1)
```
enochtsang: This loop is not intended to be indefinite, so this is dangerous.  `attemptCount > NUMBER_OF_ATTEMPTS` should go in here.

EpsilonCCS/Src/TransmitSerial.c
```diff
@@ -2,7 +2,37 @@
...
+    {
+        HAL_StatusTypeDef transmitState = HAL_UART_Transmit(&huart3, payload, bufferSize, USART_TIMEOUT_PERIOD);
+
+        if (transmitState == HAL_OK)
+        {
+            HAL_GPIO_WritePin(LED_RED_GPIO_Port, LED_RED_Pin, LED_OFF);
+            break;
```
enochtsang: Makes more sense to return here.

EpsilonCCS/Src/main.c
```diff
@@ -208,10 +227,14 @@ void SystemClock_Config(void)
     RCC_OscInitTypeDef RCC_OscInitStruct;
     RCC_ClkInitTypeDef RCC_ClkInitStruct;
 
+    /**Configure the main internal regulator output voltage
```
enochtsang: I don't think these comments will live through code generation. Unless this was generated, then ignore it.

EpsilonCCS/Src/main.c
```diff
@@ -235,24 +260,52 @@ void SystemClock_Config(void)
...
     HAL_NVIC_SetPriority(SysTick_IRQn, 15, 0);
 }
 
+/* CAN1 init function */
+static void MX_CAN1_Init(void)
+{
+
+    hcan1.Instance = CAN1;
+    hcan1.Init.Prescaler = 4;
+    hcan1.Init.Mode = CAN_MODE_LOOPBACK;
```
enochtsang: This should change to normal mode.

EpsilonCCS/Src/TransmitSerial.c
bill-luu: You mean while(attemptCount < NUMBER_OF_ATTEMPTS) right?

EpsilonCCS/Src/main.c
bill-luu: Was generated

PR #61, opened 2018-03-31
Update Orion BMS profile

Ben-Graham:


bill-luu:
image
I want to say that the other 2 relays need to be on to send messages to the Aux BMS.

bill-luu:
I'm going to let @bandakkalokka assign this to someone on electrical, as they will be a little bit more qualified to review this.

Logan-Hunter:
This is laborious to review in this format, but I can see some values that are definitely incorrect - ex max and min pack voltage seem to be 123V and 81V rather than the correct voltages, 126V and 108V. I suggest someone (Ben/Bill) come downstairs and run through the settings your changing with me ASAP.

Logan-Hunter:
Sorry, I was incorrect -min voltage is actually 81. Max is definitely 126 though

bill-luu:
@enochtsang @Logan-Hunter @Ben-Graham 
Setting the Thunderstruck Charger and the EVCC seemed to have broken the CAN messages, the issue is that we're currently using 3/4 custom flags and the those two need a custom flag each. 

Do we need to make Orion support the Thunderstruck charger also? Reading the [manuals](http://www.thunderstruck-ev.com/tsm2500-and-charge-controller.html) for the [charger](http://www.thunderstruck-ev.com/images/ThunderStruck%20TS2.5C%20ManualV1.06.pdf) & [evcc](http://www.thunderstruck-ev.com/images/EV%20Charger%20Controller%20v2.3.2b.pdf) gives me the impression that the BMS only needs to interface with the evcc, and then the charger will communicate with that.

In the event that we do, I think we can rearrange some of our current flags to match what the evcc is expecting, the manual claims that it ignores any extraneous bits, (Page 37):
image
That will free up a flag for the charger. However, also tells us to set it to zero though, so it might not work.

The last option I can think of is to drop 5 DTC messages off of the custom flags from Orion, if there are any faults that will never get triggered/we don't care about. That will leave us with two custom flags for the two of these, but I feel like that wouldn't be the best preferred though unless there are actually 5 messages we can risk dropping. What do you guys think?

bill-luu:
As per convo with Logan, we'll try enabling support for just the evcc.

bill-luu:
Let's put up a QA branch when battery is ready to go.

PR #62, opened 2018-04-29
Enoch/aux bms

enochtsang:
I think some things I would change is easier explained in actual code. Let me know if you have any questions. I know I'm not here most of the time, I'm being kept quite busy in Mexico. @bill-luu  please update this PR as you see fit. My approval is not necessary. After my PR merged, please also merge Baba's branch as you see fit. @bill-luu has final approval. 

Although there are still some things I would change, I think it's better to move onto QA.

Also my linter is failing. @bill-luu  please help me run the linter since the version on my machine seems to be different.


EpsilonAuxBMS/Middlewares/Third_Party/FreeRTOS/Source/timers.c
```diff
@@ -291,11 +291,11 @@ BaseType_t xTimerCreateTimerTask( void )
 #else
         {
             xReturn = xTaskCreate(	prvTimerTask,
-            "Tmr Svc",
-            configTIMER_TASK_STACK_DEPTH,
-            NULL,
-            ( ( UBaseType_t ) configTIMER_TASK_PRIORITY ) | portPRIVILEGE_BIT,
-            &xTimerTaskHandle );
+                                    "Tmr Svc",
```
enochtsang: This is just the linter.

bill-luu:
I would say that it isn't, the schematic doesn't name the contactors specifically. What we can do is something like:
````
#define CONTACTOR_ENABLE3_Pin GPIO_PIN_4
#define DISCHARGE_CONTACTOR_ENABLE_Pin CONTACTOR_ENABLE3
````

bill-luu:
Changes look good to me, will do another round of QA tomorrow

PR #64, opened 2018-05-03
rolling average for accel and regen values

enochtsang:


EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -140,29 +140,43 @@ void sendDriveCommandsTask(void const* arg)
...
+
     for (;;)
     {
         osDelayUntil(&prevWakeTime, DRIVE_COMMANDS_FREQ);
 
         // Read analog inputs
         if (HAL_ADC_PollForConversion(&hadc1, ADC_POLL_TIMEOUT) == HAL_OK)
         {
-            regenPercentage = ((float)HAL_ADC_GetValue(&hadc1)) / ((float)MAX_ANALOG);
+            newRegen = ((float)HAL_ADC_GetValue(&hadc1)) / ((float)MAX_ANALOG);
```
bill-luu: This will always be zero, as the float will be converted to an int later, this will apply to other similar calculations.

PR #65, opened 2018-05-04
Bill/regen accel rolling avg

bill-luu:
I managed to get constant values with the changes I did here, what do you think?


bill-luu:
I'm pretty sure something like (unsigned int)(float)/(float) would round down to the nearest int, in which case  if we divide by the maximum would always be zero.

This will let CAN report the stuff it out as a percentage value between 0-100 which might be easier to work with. Right now it's reporting a low of 5% and a high of 83%

I should double check to see how well it works with the actual motors though.

EpsilonDriverControls/Inc/DriverControls.h
```diff
@@ -32,7 +32,7 @@
 
 #define MAX_FORWARD_RPM 1000
 #define MAX_REVERSE_RPM -200
-#define NON_ZERO_THRESHOLD 0.05f
+#define NON_ZERO_THRESHOLD 0.07f
```
enochtsang: how come?

bill-luu: When I was using the knob, the lowest value alternated between 0.05 & 0.06, so I just decided to do 0.07 to be safe.

PR #81, opened 2018-07-14
Ben/orion profile

Ben-Graham:
Updated to reverse direction of current sensor BMS profile


bill-luu:
This might have been based off an older version of the profile branch - no CAN information is on here and various values that shouldn't be different, are different.

PR #96, opened 2019-01-19
Safa/update gdbflash script

safasyed:


gdbflash.sh
```diff
@@ -1,5 +1,24 @@
 #!/bin/bash
 
-cd $1/build
+if [ $# -ne 2 ]; then
+    echo "Please enter two arguments (Project name and Port number)"
```
bill-luu: We should also echo an example usage.

gdbflash.sh
```diff
@@ -1,5 +1,24 @@
...
+    exit 1
+fi
 
-arm-none-eabi-gdb $1.elf -ex "target extended-remote /dev/ttyACM$2" -ex 'monitor tpwr enable' -ex 'monitor swdp_scan' -ex 'attach 1' -ex 'load'
+
+project=$1
+port=$2
+
+if [ -d $project ]; then
+	if [ ${project: -1} == '/' ]; then
```
bill-luu: Is this a check to see if the input ends in a `/`?

I don't think that's necessary, I could be wrong but I think `cd EpsilonLights//build` does the same thing as `cd EpsilonLights/build`

gdbflash.sh
bill-luu: Also, I think instead of entering just the port number, it might be more intuitive to enter the path, e.g. `/dev/ttyACM0`. Or to provide both! Thoughts?

bill-luu: Lastly, we should make it possible to enter the arguments in any order, via flags. For example something like `./gdbflash.sh -b <black-magic-path> -p <project> `

PR #119, opened 2019-05-18
Safa/RegenLights

safasyed:
- added a filter for MotorDrive
- implemented logic for regen braking in Lights.c using motorvelocity and motorcurrentout


EpsilonLights/Inc/Lights.h
```diff
@@ -36,6 +36,10 @@
 #define AUXBMS_INPUT_STDID 0x721U
 #define STROBE_FAULT_MASK 0x01
 
+#define MOTOR_DRIVE_STDID 0x501U
+#define REGENBRAKE_INPUT_INDEX_P1 0
```
bill-luu: Name this `SET_CURRENT_INDEX`

EpsilonLights/Inc/Lights.h
```diff
@@ -36,6 +36,10 @@
...
+#define REGENBRAKE_INPUT_INDEX_P2 1
```
bill-luu: Name this `SET_VELOCITY_INDEX`

EpsilonLights/Src/Lights.c
```diff
@@ -58,6 +58,15 @@ void updateLightsTask(void const* arg)
             HAL_GPIO_WritePin(BRAKE_GPIO_Port, BRAKE_Pin, LIGHT_OFF);
         }
 
+        if (regenInputs[REGENBRAKE_INPUT_INDEX_P1] == 0 && regenInputs[REGENBRAKE_INPUT_INDEX_P2] != 0)
```
bill-luu: Combine this with the previous loop, right now the mechanical brake will be ignored because of this statement

EpsilonLights/Src/Lights.c
```diff
@@ -294,6 +303,11 @@ void HAL_CAN_RxCpltCallback(CAN_HandleTypeDef* hcan)
         auxBmsInputs[0] = msg->Data[0];
         auxBmsInputs[1] = msg->Data[1];
     }
+    else if (msg->StdId == MOTOR_DRIVE_STDID && msg->DLC == 2)
+    {
+        regenInputs[0] = msg->Data[0];
```
bill-luu: Stay consistent on whether you use the `REGENBRAKE_INPUT` macros or not, it's fine either way

EpsilonLights/Src/Lights.c
```diff
@@ -49,24 +49,15 @@ void updateLightsTask(void const* arg)
         /* UPDATE BRAKE LIGHTS */
         brakes = (driversInputs[BRAKES_INPUT_INDEX_P1] >> BRAKES_INPUT_INDEX_P2) & 1;
 
-        if (brakes)
+        if (brakes || regenInputs[SET_CURRENT_INDEX] == 0 && regenInputs[SET_VELOCITY_INDEX] != 0)
```
bill-luu: Either the logic or the labels are mismatched - regen braking is when the velocity is set to zero, and the current is set to something

bill-luu:
I was thinking about this a bit more, and there is actually a scenario in the drivers code where we step on the regen pedal, the set velocity is zero, BUT the set current will be non-zero.   
DriverControls.c:
```
        else if (regenPercentage > NON_ZERO_THRESHOLD) // Regen state
        {
            // To stop using regen braking, set motorCurrentOut to desired value and zero motorVelocityOut
            // To stop without regen braking, zero both motorCurrentOut and motorVelocityOut
            // https://tritium.com.au/includes/TRI88.004v4-Users-Manual.pdf - Section 13

            motorVelocityOut = 0;

            // Alow regen braking based on input from AuxBMS
            if (allowCharge)
            {
                motorCurrentOut = regenPercentage * REGEN_INPUT_SCALING;
            }
            else
            {
                motorCurrentOut = 0;
            }

```
In that else statement, setCurrent will be set to zero, so our pedal won't be lighting up the brake light.

We should instead check the status _of the actual pedal_, if we reach our non-zero threshold (15%), then we can set the light.

bill-luu:
The changes to the filter config and the motor drive stuff can be removed now.

PR #125, opened 2019-06-15
Mob/LapLoggingButton

safasyed:
Added a bit to driverInputs for lap button
Made Macros for LAP_PIN and LAP_GPIO_PORT to point to the CONTEXT Pin and Port
Changing context button to lap on STM was causing build issues

### DIFF
--- a/EpsilonCCS/Inc/Data/DriverControlData.h
+++ b/EpsilonCCS/Inc/Data/DriverControlData.h
@@ -28,6 +28,7 @@
 #define HORN_MASK 0x10
 #define RESET_MASK 0x20
 #define AUX_MASK 0x40
+#define LAP_MASK 0x80
 
 struct LightsInputs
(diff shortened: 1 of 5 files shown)


EpsilonCCS/Inc/Data/DriverControlData.h
```diff
@@ -57,6 +58,7 @@ struct DriverInputs
     unsigned char horn;
     unsigned char reset;
     unsigned char aux;
+    unsigned char lap;
```
bill-luu: Shouldn't this be in the `DriverControlData` struct?

bill-luu: Nevermind, github diff made it look like it was in LightsInputs....

Reddragoneer1:
Tested and verified!

PR #126, opened 2019-06-19
Avneet/NegativePackCurrent

avneetgill:
Changed pack current from unsigned int to short to account for negative values


bill-luu:
Gotta format CCS, also - this is branched off of the lap branch instead of master.

PR #144, opened 2020-04-02
AuxBMS: Remove references to hasCharge/DischargeBeenSet variables

bill-luu:
I believe that these variables to be extraneous, and the functionality associated with them can be done with the ReadPin.

I think that the issue we had during the last debugging session was caused by the AuxBMS thinking that it's calling `isContactorSet` with the wrong number of contactors. This behavior is linked to the hasCharge/DischargeBeenSet variables, so I did some digging.

These values are set by the isContactorSet call in normal operation, but when the car is nearing it's charging limits, we manually reset the variables ourselves. My guess is that the logic at:
```
else if(orionStatus.batteryVoltagesInRange)
{
   if(orionStatus.chargeContactorOverriden)
   {
        hasChargeBeenSet = 0;
    }
    ...
}
``` 

Isn't firing correctly, causing the numbering mismatch.

Looking at the usages of the variables, I think that instead of using these, I think we can directly read the sense pins to see if we need to call `isContactorSet` with 2 or 3 contactors set.

(hopefully) Fixes SFT-134

### DIFF
--- a/EpsilonAuxBMS/Src/SetContactorsTask.c
+++ b/EpsilonAuxBMS/Src/SetContactorsTask.c
@@ -54,8 +54,6 @@ void setContactorsTask(void const* arg)
 
     ContactorsSettingState state = FIRST_CHECK;
     ContactorsSettingState prevState = state;
-    uint8_t hasChargeBeenSet = 0;
-    uint8_t hasDischargeBeenSet = 0;
 
     for (;;)
     {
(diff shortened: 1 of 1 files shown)


EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -162,30 +153,31 @@ void setContactorsTask(void const* arg)
...
+                    {
+                        HAL_GPIO_WritePin(DISCHARGE_ENABLE_GPIO_Port, DISCHARGE_ENABLE_Pin, GPIO_PIN_SET);
+                    }
                 }
 
                 uint8_t commonSense = !HAL_GPIO_ReadPin(COMMON_SENSE_GPIO_Port, COMMON_SENSE_Pin);
-                uint8_t dischargeSense = !HAL_GPIO_ReadPin(DISCHARGE_SENSE_GPIO_Port, DISCHARGE_SENSE_Pin);
 
-                if (!commonSense || (hasDischargeBeenSet && !dischargeSense && !orionStatus.dischargeContactorOverriden))
+                if (!commonSense || (!dischargeSense && !orionStatus.dischargeContactorOverriden))
```
badeniran, line 180: This may not work under normal operation. The sequence is Common -> Charge, so at this point, discharge wouldn't be set, so I think the contactors would get disconnected every time. If think we would have to check that the previous state was `DISCHARGE_ENABLE_CHECK` or `DONE`?

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -243,30 +227,31 @@ void setContactorsTask(void const* arg)
...
+                    {
+                        HAL_GPIO_WritePin(CHARGE_ENABLE_GPIO_Port, CHARGE_ENABLE_Pin, GPIO_PIN_SET);
+                    }
                 }
 
                 uint8_t commonSense = !HAL_GPIO_ReadPin(COMMON_SENSE_GPIO_Port, COMMON_SENSE_Pin);
-                uint8_t chargeSense = !HAL_GPIO_ReadPin(CHARGE_SENSE_GPIO_Port, CHARGE_SENSE_Pin);
 
-                if (!commonSense || (hasChargeBeenSet && !chargeSense && !orionStatus.chargeContactorOverriden))
+                if (!commonSense || (!chargeSense && !orionStatus.chargeContactorOverriden))
```
badeniran, line 254: Same as above

EpsilonAuxBMS/Src/SetContactorsTask.c
bill-luu: That would work for DONE, but not if the discharge check finished with the discharge contactor not closing.... maybe we do need to keep track of that?

badeniran: Hmmm. The other question is do we want to disconnect the contactors if charge or discharge is open but should be closed?

bill-luu: Don't we do that in done?

badeniran: Yeah,  maybe we only do it in done?

duan-le:
It looks good to me!

PR #146, opened 2020-05-27
SFT-117 Smooth jumps in current in Driver Controls

badeniran:
During FSGP 2019, many issues with motors tripping when switching 0 current out to high current out.
IIR Low Pass Filter (system that is resistant to change) implemented to solve this issue.

### DIFF
--- a/EpsilonDriverControls/Inc/DriverControls.h
+++ b/EpsilonDriverControls/Inc/DriverControls.h
@@ -37,6 +37,7 @@
 #define MAX_PEDAL_THRESHOLD 0.71f
 #define MAX_ANALOG 4095 // 12bit ADC (2^12)
 #define REGEN_INPUT_SCALING 0.5f
+#define MOTOR_CURRENT_SMOOTHING_FACTOR 0.15f // Smooth current output to prevent big jumps (0-1)
 
 #define AUXBMS_INPUT_STDID 0x721U
 
--- a/EpsilonDriverControls/Src/DriverControls.c
+++ b/EpsilonDriverControls/Src/DriverControls.c
@@ -52,6 +52,26 @@ float calculateMotorCurrent(float accelPercentage)
     }
 }
 
+float lowPassFilter(float presentMotorCurrent, float prevMotorCurrent)
+{
+    // Essentially a simple IIR low pass filter, which is a system that resists change.
+    // The purpose is to smooth the output current so that there aren't any big jumps that could cause the motors to trip.
+    // The MOTOR_CURRENT_SMOOTHING_FACTOR determines how resistant the low pass filter is to change
+    return prevMotorCurrent + MOTOR_CURRENT_SMOOTHING_FACTOR * (presentMotorCurrent - prevMotorCurrent);
+}
+
+float calculateAccelMotorCurrent(float accelPercentage, float prevMotorCurrent)
+{
+    return lowPassFilter(calculateMotorCurrent(accelPercentage), prevMotorCurrent);
+}
+
+float calculateRegenMotorCurrent(float regenPercentage, float prevMotorCurrent)
+{
+    // Scale presentMotorCurrent by REGEN_INPUT_SCALING because regen uses less current than normal acceleration. Motors will trip if current is greater.
+    float presentMotorCurrent = calculateMotorCurrent(regenPercentage) * REGEN_INPUT_SCALING ;
+    return lowPassFilter(presentMotorCurrent, prevMotorCurrent);
+}
+
 void sendHeartbeatTask(void const* arg)
 {
     uint32_t prevWakeTime = osKernelSysTick();
@@ -237,9 +257,7 @@ void sendDriveCommandsTask(void const* arg)
             // Alow regen braking based on input from AuxBMS
             if (allowCharge)
             {
-                // Regen needs to be scaled more to avoid motor trips
-
-                motorCurrentOut = calculateMotorCurrent(regenPercentage) * REGEN_INPUT_SCALING;
+                motorCurrentOut = calculateRegenMotorCurrent(regenPercentage, motorCurrentOut);
             }
             else
             {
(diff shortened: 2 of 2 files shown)


badeniran:
A smoothing factor of 0.15 was pretty arbitrary. It could potentially be bigger or may need to be smaller, but testing will be needed to find a value that gives us what we want

EpsilonDriverControls/Inc/DriverControls.h
```diff
@@ -36,7 +36,7 @@
 #define NON_ZERO_THRESHOLD 0.17f
 #define MAX_PEDAL_THRESHOLD 0.71f
 #define MAX_ANALOG 4095 // 12bit ADC (2^12)
-#define REGEN_INPUT_SCALING 0.5f
```
bill-luu, line 39: This is still removed

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -256,13 +256,17 @@ void sendDriveCommandsTask(void const* arg)
         {
             HAL_GPIO_TogglePin(LED_BLUE_GPIO_Port, LED_BLUE_Pin);
             motorVelocityOut = MAX_FORWARD_RPM;
-            motorCurrentOut = calculateMotorCurrent(accelPercentage);
+
+            // Smoothing output current so that there aren't any big jumps that could cause the motors to trip. Essentially a simple IIR low pass filter
+            motorCurrentOut += MOTOR_CURRENT_SMOOTHING_FACTOR * (calculateMotorCurrent(accelPercentage) - motorCurrentOut);
```
bill-luu, line 261: I think the logic of smoothing the current should be put into `calculcateMotorCurrent`, so that this function just has to call:
`motorCurrentOut = calculateMotorCurrent(motorCurrentOut)`

EpsilonDriverControls/Inc/DriverControls.h
badeniran: Oops, forgot to save the header file haha

EpsilonDriverControls/Src/DriverControls.c
badeniran: This is a bit more complicated with the regen calculation because the regen scale factor is applied to the return value of calculateMotorCurrent, then the smoothing is done

badeniran: This is a bit more complicated with regen because the regen scaling is applied to the return value of `calculateMotorCurrent` then the smoothing is done. Would it be alright if I created two functions called `calculateAccelMotorCurrent` and `calculateRegenMotorCurrent`, then make those functions call `calculateMotorCurrent` as it is, then have the smoothing done in `calculateAccelMotorCurrent` and `calculateRegenMotorCurrent` with the different functionality needed in each?

badeniran: One more thing, when the percentage is less than the non-zero threshold, `calculateMotorCurrent` returns 0, but with the smoothing, it wouldn't instantly go to 0, but would slowly go to 0 if it was at a non-zero value before. Is this okay, or should I add the functionality so that if 0 is returned, then smoothing is skipped and 0 is assigned?

bill-luu: That sounds good! As for the returning 0, it's OK if it smooths over.

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -52,6 +52,19 @@ float calculateMotorCurrent(float accelPercentage)
...
+    // Smoothing output current so that there aren't any big jumps that could cause the motors to trip. Essentially a simple IIR low pass filter
+    return prevMotorCurrent + MOTOR_CURRENT_SMOOTHING_FACTOR * (calculateMotorCurrent(accelPercentage) - prevMotorCurrent);
+}
+
+float calculateRegenMotorCurrent(float regenPercentage, float prevMotorCurrent)
+{
+    // Smoothing output current so that there aren't any big jumps that could cause the motors to trip. Essentially a simple IIR low pass filter.
+    // Scaled by REGEN_INPUT_SCALING because regen uses less current than normal acceleration. Motors will trip if current is greater.
+    return prevMotorCurrent + MOTOR_CURRENT_SMOOTHING_FACTOR * (calculateMotorCurrent(regenPercentage) * REGEN_INPUT_SCALING - prevMotorCurrent);
+}
```
bill-luu, line 66: I think we can extract the smoothing function out again:
```
float smoothCurrent(float prevMotorCurrent, float presentMotorCurrent)
{
    // Smoothing output current so that there aren't any big jumps that could cause the motors to trip. Essentially a simple IIR low pass filter
     return prevMotorCurrent + MOTOR_CURRENT_SMOOTHING_FACTOR * (presentMotorCurrent- prevMotorCurrent);
}
float calculateAccelMotorCurrent(float accelPercentage, float prevMotorCurrent)
{
    return smoothCurrent(prevMotorCurrent, calculateMotorCurrent(accelPercentage));
}

float calculateRegenMotorCurrent(float regenPercentage, float prevMotorCurrent)
{
    float presentRegenCurrent = calculateMotorCurrent(regenPercentage) * REGEN_INPUT_SCALING;
    return smoothCurrent(prevMotorCurrent, presentRegenCurrent);
}
```

bill-luu:
Does `smoothCurrent` make enough sense as a function name to a new reader, or should we call it something like `lowPassFilter`?

badeniran:
> Does `smoothCurrent` make enough sense as a function name to a new reader, or should we call it something like `lowPassFilter`?

That might be a good idea. I guess I could also add a quick description as to what a low pass filter is

avneetgill:
> > Does `smoothCurrent` make enough sense as a function name to a new reader, or should we call it something like `lowPassFilter`?
> 
> That might be a good idea. I guess I could also add a quick description as to what a low pass filter is

PR #148, opened 2020-05-30
SFT-137 Stop car before changing forward/reverse

badeniran:
The old code had the potential of the car jerking to try and change directions if the driver switched the direction knob while driving. This fixes it by causing the car to slow down to a stop before switching the direction of the car.


badeniran:
One thing is whether we should have a different threshold instead of 0 to determine whether the car has stopped. I'm not sure since I haven't seen it, but do the motors actually report 0 when the car is stopped or just some value close to 0?

badeniran:
Additionally, would it be fine to just just one motor to determine direction since the assumption is that they're both going in the same direction anyways? Or would it be "safer" to use both in case something goes catastrophically wrong haha

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -362,6 +403,16 @@ void HAL_CAN_RxCpltCallback(CAN_HandleTypeDef* hcan)
         auxBmsInputs[1] = msg->Data[1];
     }
 
+    if (msg->StdId == M0_VELOCITY_INPUT_STDID && msg->DLC == 8)
+    {
+        motorVelocityInput[0] = arrayToFloat(&(msg->Data[4]));
```
bill-luu, line 408: Starting from Data[4], isn't it `VehicleVelocity`, not `motorVelocity`?

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -362,6 +403,16 @@ void HAL_CAN_RxCpltCallback(CAN_HandleTypeDef* hcan)
...
+    }
+
+    if (msg->StdId == M1_VELOCITY_INPUT_STDID && msg->DLC == 8)
+    {
+        motorVelocityInput[1] = arrayToFloat(&(msg->Data[4]));
```
bill-luu, line 413: Same as above

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -232,9 +256,15 @@ void sendDriveCommandsTask(void const* arg)
...
-        forward = !HAL_GPIO_ReadPin(FORWARD_GPIO_Port, FORWARD_Pin); // `!` for active low
-        reverse = !HAL_GPIO_ReadPin(REVERSE_GPIO_Port, REVERSE_Pin);
+        // Determine vehicle direction
+        VehicleDirection vehicleDirection = determineDirection();
+
+        // Determine drive commands
+        // Forward and reverse are only asserted if the respective GPIO pin is asserted and the car is moving in that direction or is stationary
+        // If the GPIO pin is asserted, but the car is moving in the opposite direction, the output velocity and current will be set to 0, making
+        // the car coast to a stop
+        forward = !HAL_GPIO_ReadPin(FORWARD_GPIO_Port, FORWARD_Pin) && (vehicleDirection == FORWARD || vehicleDirection == STATIONARY); // `!` for active low
```
bill-luu, line 266: `!HAL_GPIO_ReadPin(FORWARD_GPIO_Port, FORWARD_Pin) && (vehicleDirection == FORWARD || vehicleDirection == STATIONARY)`
I'd put that in a function, this boolean statement is getting unweildy.

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -232,9 +256,15 @@ void sendDriveCommandsTask(void const* arg)
...
+        reverse = !HAL_GPIO_ReadPin(REVERSE_GPIO_Port, REVERSE_Pin) && (vehicleDirection == REVERSE || vehicleDirection == STATIONARY);
```
bill-luu, line 267: Same as above

EpsilonDriverControls/Inc/DriverControls.h
```diff
@@ -50,6 +53,9 @@ extern ADC_HandleTypeDef hadc2;
 extern osMessageQId canQueue;
 extern osPoolId canPool;
 extern uint8_t auxBmsInputs[2];
+extern float   motorVelocityInput[2];
```
bill-luu, line 56: I think 2 separate floats makes more sense, given these are for 2 different motors and are not part of the same input.

bill-luu:
For consideration, see [how](https://github.com/UCSolarCarTeam/Delta-CCS-Software/blob/master/DriverControl/DriverControl.cpp) delta handled directions.

badeniran:
> Partial review.
> Are there any ways this could go wrong if suddenly the motorControllers stopped reporting information? It seems like the motorVelocityInputs aren't being reset, so it would hold the last value sent.

Yeah there would only be an issue if the velocity that is held is 0 because the car will be interpreted as stationary and we would be back to the same problem. One thing is that we could implement a `motorControllersRecentlyReceived` field and only adjust if information had recently been received and perhaps add an error for driver controls? It may not be necessary to add the error field since I assume that CCS would also detect that we aren't receiving motor controller information, but we could add it just to be safe (could be a driver controls only thing).

bill-luu:
Actually, I don't think it will be that big a deal, I can't imagine any situation where we aren't able to receive CAN messages from the motor controller, but are somehow able to send messages to it - it's the same CAN wire lol

EpsilonDriverControls/Inc/DriverControls.h
```diff
@@ -44,6 +44,9 @@
 #define M0_VELOCITY_INPUT_STDID 0x403U
 #define M1_VELOCITY_INPUT_STDID 0x423U
 
+#define SAFE_VEHICLE_VELOCITY_TO_GO_FORWARD 0.0f // Will probably be 0 or a small negative number
+#define SAFE_VEHICLE_VELOCITY_TO_GO_REVERSE 0.0f // Will probably be 0 or a small positive number
```
bill-luu, line 48: Maybe give a bit of buffer, I don't know how it will react on slow turns from rest - eg if you turn right, the left motor goes forward but the right one might stay staionary or move backward a bit?

badeniran: Do you have an idea of a good number? Maybe like +/- 10 km/h?

PR #149, opened 2020-05-30
User/jessie duan/sft 100

JessieG123:
Created functions for each of the cases and make some of the longer IF statements more readable by making functions as well.


EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -47,16 +47,41 @@ void resetSenseArray(uint32_t senses[]);
...
+void chargeEnableCheck();
+void dischargeEnableCheck();
+void done();
+void blocked();
+
+// Global pointers to be used by the state logic functions
+ContactorsSettingState *ptrState;
+ContactorsSettingState *ptrPrevState;
+uint8_t *ptrHasChargeBeenSet;
+uint8_t *ptrHasDischargeBeenSet;
```
badeniran: Instead of making these pointers, you could keep them as global variables and do the one time assignment to them below.

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -47,16 +47,41 @@ void resetSenseArray(uint32_t senses[]);
...
+
 void setContactorsTask(void const* arg)
 {
-    // One time osDelayUntil initialization
-    uint32_t prevWakeTime = osKernelSysTick();
-
     ContactorsSettingState state = FIRST_CHECK;
     ContactorsSettingState prevState = state;
     uint8_t hasChargeBeenSet = 0;
     uint8_t hasDischargeBeenSet = 0;
```
badeniran, line 75: Replace these lines with the one time assignment to the global variables

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -545,3 +289,308 @@ void disconnectContactors(uint8_t isContactorError)
...
+        *ptrState = SHUTDOWN;
+        return;
+    }
+    else if (!commonSense)
+    {
+        disconnectContactors(1);
+        *ptrState = CONTACTOR_DISCONNECTED;
+    }
+
+    if (orionStatus.allowCharge && orionStatus.chargeContactorOverriden && !chargeSense)
```
badeniran, line 505: Could we simplify this if statement as well?

badeniran: Maybe assign the condition to a variable called `canTurnOnChargeContactor`

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -545,3 +289,308 @@ void disconnectContactors(uint8_t isContactorError)
...
+        osMutexRelease(orionStatus.orionStatusMutex);
+
+        isChargeTurningOn = 1;
+        *ptrHasChargeBeenSet = 0;
+        // Turn on charge Contactor
+        HAL_GPIO_WritePin(CHARGE_ENABLE_GPIO_Port, CHARGE_ENABLE_Pin, GPIO_PIN_SET);
+        *ptrPrevState = *ptrState;
+        *ptrState = CHARGE_ENABLE_CHECK;
+    }
+    else if (orionStatus.allowDischarge && orionStatus.dischargeContactorOverriden && !dischargeSense)
```
badeniran, line 522: Same as above

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -545,3 +289,308 @@ void disconnectContactors(uint8_t isContactorError)
...
+        *ptrHasDischargeBeenSet = 0;
+        HAL_GPIO_WritePin(DISCHARGE_ENABLE_GPIO_Port, DISCHARGE_ENABLE_Pin, GPIO_PIN_SET);
+        *ptrPrevState = *ptrState;
+        *ptrState = DISCHARGE_ENABLE_CHECK;
+    }
+
+
+    // If charge contactor should be on, but isn't, disconnect
+    // or if discharge contactor should be on, but isn't, disconnect
+    if ((orionStatus.allowCharge && !chargeSense && !isChargeTurningOn && !orionStatus.chargeContactorOverriden) ||
```
badeniran, line 542: Same as above, but two variables. One for each condition on each side of the `||`

EpsilonAuxBMS/Src/SetContactorsTask.c
bill-luu: I was the one that recommended pointers, but looking at it now I think that probably isn't the best idea :/
What do you think of making make an entire struct collecting these 4 fields together, then you could do like:
```
myStruct A(struct ) {
//blah
return aStruct
}
switch(myStuct.state){
case A:
 myStruct = A(myStruct)
}
```

badeniran: Wooow Bill tsk tsk. Yeah it was a bit messy to have all those asterisks everywhere.
I think what may be better is if the argument to the functions was a pointer to a struct, then they can use the `->` to dereference the values. And it will eliminate the need to keep assigning to the struct and returning a struct in each function. So like:

`void A(struct* aStruct){
// blah
if(aStruct->val) doSomething;
aStruct->state = B;
}

void setContactors(){
  // blah
  struct myStruct;
  // blah
  switch(myStruct.state){
  case A:
    A(&myStruct);
  case B:
    //blah
  }
`

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -67,6 +67,18 @@ const uint8_t isDischargeFalsePositive(const uint8_t dischargeSense);
...
 const uint8_t isChargeFalsePositive(const uint8_t chargeSense);
 
+// Checks if discharge contactor can be turn on
+const uint8_t canTurnOnDishargeContactor(const uint8_t dischargeSense);
+
+// Checks if charge contactor can be turn on
+const uint8_t canTurnOnChargeContactor(const uint8_t chargeSense);
+
+// Checks if the discharge contactor should be on, but isn't
+const uint8_t isDischargeContactorShouldBeOnButNot(const uint8_t dischargeSense, const uint8_t isdischargeTurningOn);
```
badeniran, line 77: Maybe take out the `is` in the name? It's a bit confusing. I was trying to think of better names, but it's sort of a hard function to name haha

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -67,6 +67,18 @@ const uint8_t isDischargeFalsePositive(const uint8_t dischargeSense);
...
+
+// Checks if the charge contactor should be on, but isn't
+const uint8_t isChargeContactorShouldBeOnButNot(const uint8_t chargeSense, const uint8_t isChargeTurningOn);
```
badeniran, line 80: Same thing here

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -545,3 +295,329 @@ void disconnectContactors(uint8_t isContactorError)
...
+{
+    return (orionStatus.allowDischarge && orionStatus.dischargeContactorOverriden && !dischargeSense);
+}
+
+const uint8_t canTurnOnChargeContactor(const uint8_t chargeSense)
+{
+    return (orionStatus.allowCharge && orionStatus.chargeContactorOverriden && !chargeSense);
+}
+
+const uint8_t dischargeContactorShouldBeOnButNot(const uint8_t dischargeSense, const uint8_t isDischargeTurningOn)
```
bill-luu, line 615: Now that I think of it, both this function and `isDischargeFalsePositive` both check if the contactor disconnected, this one just checks it during the DONE state and the other one checks if the contactor disconnected during the CHECK/BLOCK phases. The names should probably reflect it, in a way that makes sense to someone that might not necessarily know the states in the code.
Maybe `dischargeDisconnectedDuringContactorSetting` for `falsePositive` and `dischageDisconnectedUnexpectedly` for `shouldBeOnButNot`? still sounds a little janky haha

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -545,3 +295,329 @@ void disconnectContactors(uint8_t isContactorError)
...
+        if (osMutexWait(orionStatus.orionStatusMutex, 0) != osOK)
+        {
+            return;
+        }
+
+        orionStatus.dischargeContactorOverriden = 0;
+        osMutexRelease(orionStatus.orionStatusMutex);
+
+        isDischargeTurningOn = 1;
+        toAlsoBeNamed->hasDischargeBeenSet = 0;
```
bill-luu, line 539: @badeniran This might be a future issue, but I have a hunch that instead of having a `isDischargeTurningOn`, we can replace that with a `!hasDischargeBeenSet` usage.

Going further, and this will need testing but I think if we move the checking if contactors disconnected from the end of `blocked` to the beginning, like the other functions, we should be able to simplify the checking logic in `..shouldBeOnButNot` functions

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -545,3 +295,329 @@ void disconnectContactors(uint8_t isContactorError)
...
+        HAL_GPIO_WritePin(DISCHARGE_ENABLE_GPIO_Port, DISCHARGE_ENABLE_Pin, GPIO_PIN_SET);
+        toAlsoBeNamed->prevState = toAlsoBeNamed->state;
+        toAlsoBeNamed->state = DISCHARGE_ENABLE_CHECK;
+    }
+
+
+    // If charge contactor should be on, but isn't, disconnect
+    // or if discharge contactor should be on, but isn't, disconnect
+    if (chargeContactorShouldBeOnButNot(chargeSense, isChargeTurningOn) || dischargeContactorShouldBeOnButNot(dischargeSense, isDischargeTurningOn))
```
bill-luu, line 548: This should get seperated into two lines
```
    if (chargeContactorShouldBeOnButNot(chargeSense, isChargeTurningOn)
    || dischargeContactorShouldBeOnButNot(dischargeSense, isDischargeTurningOn))
```

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -35,6 +35,14 @@ typedef enum ContactorsSettingState
     SHUTDOWN
 } ContactorsSettingState;
 
+typedef struct ToBeNamed
```
bill-luu, line 38: idk maybe like `SetContactorStateInfo`?

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -545,3 +295,329 @@ void disconnectContactors(uint8_t isContactorError)
...
+        orionStatus.chargeContactorOverriden = 0;
+        orionStatus.dischargeContactorOverriden = 0;
+
+        osMutexRelease(orionStatus.orionStatusMutex);
+        toAlsoBeNamed->state = FIRST_CHECK;
+    }
+}
+
+const uint8_t isDischargeFalsePositive(const uint8_t dischargeSense, uint8_t hasDischargeBeenSet)
+{
```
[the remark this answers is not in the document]
badeniran: I think this is more discharge should be on, but isn't on. But as bill mentions bellow, it's similar to the other function and needs to be differentiated.

EpsilonAuxBMS/Src/SetContactorsTask.c
badeniran: I think the names you've suggested make sense. They're janky, but at least there's no ambiguity haha

EpsilonAuxBMS/Src/SetContactorsTask.c
badeniran: I wish we had documented all the things we tried haha, but I think we tried doing the checking logic in the beginning and there were issues. I'll have a look at your first concern though. You might be right, but it may affect `chargeContactorShouldBeOnButNot ` and `dischargeContactorShouldBeOnButNot `

bill-luu: Talked with Baba, we'll make a separate issue for this, dw about addressing it here.

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -47,344 +55,89 @@ void resetSenseArray(uint32_t senses[]);
...
-                    state = CHARGE_ENABLE_CHECK;
-                }
-
-
-
-                break;
+                dischargeEnableCheck(&toAlsoBeNamed);
             }
 
             case DONE:
```
bill-luu, line 125: I guess this is a good time to rename this state, `DONE` made sense back then, but now it doesn't represent any real finality in the Aux BMS. What kind of name should we give this?

badeniran: Maybe `IDLE` since this is a state that the Aux BMS waits around for things to change?

EpsilonAuxBMS/Src/SetContactorsTask.c
```diff
@@ -47,344 +55,89 @@ void resetSenseArray(uint32_t senses[]);
...
+const uint8_t canTurnOnChargeContactor(const uint8_t chargeSense);
+
+// Checks if the discharge contactor should be on, but isn't
+const uint8_t dischargeDisconnectedUnexpectedly(const uint8_t dischargeSense, const uint8_t isdischargeTurningOn);
+
+// Checks if the charge contactor should be on, but isn't
+const uint8_t chargeDisconnectedUnexpectedly(const uint8_t chargeSense, const uint8_t isChargeTurningOn);
+
+//First check before setting states in case battery voltages are not in range, start up sequence is not done and makes sure the state is not disconnected
+const uint8_t safeToSetContactor(ContactorsSettingState state);
```
bill-luu, line 85: Can we move these function declarations inside the header? That way unit tests can reference them later.

bill-luu:
Still should move these into different files as mentioned earlier.

EpsilonAuxBMS/Inc/SetContactorsTask.h
```diff
@@ -5,6 +5,38 @@
...
+const uint8_t canTurnOnChargeContactor(const uint8_t chargeSense);
+
+// Checks if the discharge contactor should be on, but isn't
+const uint8_t dischargeDisconnectedUnexpectedly(const uint8_t dischargeSense, const uint8_t isdischargeTurningOn);
+
+// Checks if the charge contactor should be on, but isn't
+const uint8_t chargeDisconnectedUnexpectedly(const uint8_t chargeSense, const uint8_t isChargeTurningOn);
+
+//First check before setting states in case battery voltages are not in range, start up sequence is not done and makes sure the state is not disconnected
+const uint8_t safeToSetContactor(ContactorsSettingState state);
```
badeniran, line 42: @bill-luu Just to confirm. Did you want the prototypes and implementation in completely new source and header files?

bill-luu: States should be in different files, and the other functions should be moved accordingly to who uses it.

badeniran: @bill-luu By separate files, do you mean that there should be a file for each state? Or just one file that has all the states and then the helper functions in a separate file?

EpsilonAuxBMS/Src/SetContactor/StateFunctions.c
```diff
@@ -0,0 +1,357 @@
...
+
+const uint8_t chargeDisconnectedUnexpectedly(const uint8_t chargeSense, const uint8_t isChargeTurningOn)
+{
+    return (orionStatus.allowCharge && !chargeSense && !isChargeTurningOn && !orionStatus.chargeContactorOverriden);
+}
+
+const uint8_t safeToSetContactor(ContactorsSettingState state)
+{
+    return (!orionStatus.batteryVoltagesInRange && !auxStatus.startUpSequenceDone && state != CONTACTOR_DISCONNECTED);
+}
```
badeniran, line 357: It might be a good idea to move these functions to their own file. @bill-luu What do you think?

badeniran: Eh, it's not a big deal unless Bill cares

EpsilonAuxBMS/Inc/SetContactorsTask.h
bill-luu: I guess I forgot to answer this question

EpsilonAuxBMS/Src/SetContactor/StateFunctions.c
bill-luu: If you guys aren't too tired of this, if we put each state in it's own file, then functions that are exclusively used in that a state can go into that file. (e.g `chargeDisconnectedUnexpectedly` and `dischargeDisconnectedUnexpectedly` can go in `idle.c`), and common ones can stay here. 

I notice that `safeToSetContactor` is the only function _not_ being used by any of the states but instead just in `SetContactorsTask.c`. If thats the case it's ok to just move it over there!

EpsilonAuxBMS/Src/SetContactor/StateFunctions.c
```diff
@@ -0,0 +1,357 @@
...
+void blocked(SetContactorStateInfo* stateInfo)
+{
+    const uint8_t commonSense = !HAL_GPIO_ReadPin(COMMON_SENSE_GPIO_Port, COMMON_SENSE_Pin);
+    const uint8_t chargeSense = !HAL_GPIO_ReadPin(CHARGE_SENSE_GPIO_Port, CHARGE_SENSE_Pin);
+    const uint8_t dischargeSense = !HAL_GPIO_ReadPin(DISCHARGE_SENSE_GPIO_Port, DISCHARGE_SENSE_Pin);
+
+    // If orion isn't overiding the contactors, but they are supposed to have been set, but aren't
+    // , this is an error condition and all contactors should be disconnected.
+    if (stateInfo->prevState > COMMON_ENABLE_CHECK &&
+            (!commonSense || chargeDisconnectedDuringContactorSetting(chargeSense, stateInfo->hasChargeBeenSet) ||
```
bill-luu, line 293: Each of the conditions should be on a different line:
```
            (!commonSense ||
             chargeDisconnectedDuringContactorSetting(chargeSense, stateInfo->hasChargeBeenSet) ||
             dischargeDisconnectedDuringContactorSetting(dischargeSense, stateInfo->hasDischargeBeenSet)))
```

EpsilonAuxBMS/Src/SetContactor/CheckContactorSet.c
```diff
@@ -1,3 +1,4 @@
+#pragma once
```
bill-luu, line 1: Include guards go in the header, not the source files.

bill-luu:
1 more small thing, can we move the 5 states into a folder, so like `SetContactor/States/idle.c`? I promise that's the last design change

badeniran:
Just fix the merge conflicts

EpsilonAuxBMS/Inc/SetContactor/CheckContactorSet.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+#include "AuxBmsTasks.h"
+#include "stm32f4xx_hal_adc.h"
```
bill-luu, line 3: Oh, for unit test reasons, you'll need to also include stm32f4xx_hal.h before this.

bill-luu:
image

PR #150, opened 2020-06-05
Changed code to match old lights board code

badeniran:


badeniran:
The build will fail till we merge Avneet's branch into this one.

PR #152, opened 2020-06-06
Add CMock to the Embedded Repo

bill-luu:
CMock is a mock generator that lets us spoof headers and functions that we aren't testing.

This PR introduces it to the embedded repo, and makes an example test (`SetContactorsTest.c`) in the AuxBMS unit tests.

Don't worry about the actual content of the tests in `SetContactorsTask.c` - that's a later problem. This PR is just a proof of concept on how to use the cmock expects and the like.


bill-luu:
To get the submodules, you will need to run
`git submodule update --init`

UnityTestingTool/AuxBMS/SetContactorsTaskTest.c
```diff
@@ -0,0 +1,20 @@
...
+    RUN_TEST(test_disconnectContactors);
+}
+
+void test_disconnectContactors()
+{
+    HAL_GPIO_WritePin_Expect(HV_ENABLE_GPIO_Port, HV_ENABLE_Pin, GPIO_PIN_RESET);
+    HAL_GPIO_WritePin_Expect(DISCHARGE_ENABLE_GPIO_Port, DISCHARGE_ENABLE_Pin, GPIO_PIN_RESET);
+    HAL_GPIO_WritePin_Expect(CHARGE_ENABLE_GPIO_Port, CHARGE_ENABLE_Pin, GPIO_PIN_RESET);
+    HAL_GPIO_WritePin_Expect(COMMON_ENABLE_GPIO_Port, COMMON_ENABLE_Pin, GPIO_PIN_RESET);
+    disconnectContactors(0);
```
badeniran, line 19: Does this work? Because `disconnectContactors` doesn't have a prototype in `SetContactorsTask.h` the prototype is in `SetContactorsTask.c`

badeniran: Also, ran the test that line commented out and it still passes. Also tried running it with a value of 1 being passed in and it fails with this message:
`main.c:10:test_disconnectContactors:FAIL:Function osMutexWait.  Called more times than expected.`

bill-luu: That explains why I was getting an implicit declaration warning! I'll move it to the header.

badeniran: Oh with my second comment, I get what's happening now! Ignore the first part of me commenting out, but the changing of the value to 1 is interesting. Do we need to specify how many times we expect `osMutexWait` to be called?

bill-luu: Oooh, I see what you mean. I think the way that we could test it without changing disconnectContactors is making 2 tests - one with a 0 arg, and one with a 1 arg, to cover both paths.

badeniran: Okay sounds good!

bill-luu: Side note - we only call `disconnectContactors(0)` when we are about to enter a shutdown state. Would it be unreasonable to call `disconnectContactors(1)` instead and always execute the auxStatus updates?

badeniran: I think that's fair. If `disconnectContactors(1)` works, then we've tested the functionality of `disconnectContactors(0)`. If we want to test the 0 arg, then we could make sure the auxStatus updates don't happen, but I'm not sure what the benefit of doing that would be

bill-luu: What I mean is more in the actual functionality, I feel like we could just always have the auxStatus be updated, and have a function `disconnectContactors()`=`disconnectContactors(1)`, or even two functions - `disconnectContactors` and one for the auxStatus update.

badeniran: Ah I see. I think the reason was because tripping in  `UpdateCharge` wasn't due to a contactor error so in that case we didn't want to falsely say that the car tripped due to a contactor error. So I would say the second option is probably a good idea. Then in `SetContactors` we can wrap both functions in one function

PR #157, opened 2020-07-11
Move a few variables into the for loop

duan-le:


EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -206,12 +206,6 @@ void sendDriverTask(void const* arg)
 void sendDriveCommandsTask(void const* arg)
 {
     uint32_t prevWakeTime = osKernelSysTick();
-    float newRegen = 0;
-    float newAccel = 0;
-    uint8_t forward = 0;
-    uint8_t reverse = 0;
-    uint8_t brake = 0;
-    uint8_t reset = 0;
     float busCurrentOut = 1.0f; // Percentage 0 -1 always 100%
```
bill-luu, line 209: I think this value never changes at all, so it can maybe be changed to a macro instead?

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -206,12 +206,6 @@ void sendDriverTask(void const* arg)
...
     float motorVelocityOut = 0; // RPM
```
bill-luu, line 210: This can be moved in too I think, as well as `dataToSendFloat`

EpsilonDriverControls/Src/DriverControls.c
```diff
@@ -206,11 +207,7 @@ void sendDriverTask(void const* arg)
 void sendDriveCommandsTask(void const* arg)
 {
     uint32_t prevWakeTime = osKernelSysTick();
-    float busCurrentOut = 1.0f; // Percentage 0 -1 always 100%
-    float motorVelocityOut = 0; // RPM
-    float motorCurrentOut = 0.0f; // Percentage 0 - 1
```
bill-luu, line 217: `motorCurrentOut` needs to stay outside, because the lines:
`motorCurrentOut = calculateRegenMotorCurrent(regenPercentage, motorCurrentOut);`
Require a saved value of motorCurrentOut to do their calculations.

duan-le:
I think I broke stuff. I'm going to remake it.

PR #161, opened 2020-07-24
Porting over lights board code to new firmware and new board

badeniran:
Creating a new PR because of weird git things with the old one


bill-luu:
Add the jira numbers to the commits

PR #165, opened 2020-08-07
SFT-158 Create make.sh for Driver Controls 

JessieG123:
`make clean` source code and test code when `./make.sh clean` is executed.
Build source code
Build test code
Run test code

### DIFF
--- a/EpsilonDriverControls/make.sh
+++ b/EpsilonDriverControls/make.sh
@@ -5,5 +5,5 @@ then
     exit
 fi
 
-(cd ../ && ./format.sh EpsilonAuxBMS)
+(cd ../ && ./format.sh EpsilonDriverControls)
 (make -j4 && cd Tests/ && make)
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_script:
 
 script:
     - (cd EpsilonLights && make)
-    - (cd EpsilonDriverControls && make)
+    - (cd EpsilonDriverControls && ./make.sh)
     - (cd EpsilonCCS && make)
     - (cd EpsilonAuxBMS && ./make.sh)
     - "! (./format.sh EpsilonCCS | grep Formatted)"


EpsilonDriverControls/make.sh
```diff
@@ -0,0 +1,11 @@
+if [ "$1" = "clean" ]
+then
+    (cd ../EpsilonDriverControls/ && make clean)
+    (cd Tests/ && make clean)
+    exit
+fi
+
+#!/bin/bash
```
bill-luu, line 8: Shebangs need to be at the top of the file.
https://en.wikipedia.org/wiki/Shebang_(Unix)

EpsilonDriverControls/make.sh
```diff
@@ -0,0 +1,11 @@
...
+if [ "$1" = "clean" ]
+then
+    (cd ../EpsilonDriverControls/ && make clean)
```
bill-luu, line 4: Wait, doesn't `cd ../EpsilonDriverControls` just take you to the same spot?

JessieG123: Oh wait yea true.

JessieG123: But then when you're making the source code why do you browse to the folder when you're already in it when running make.sh?

EpsilonDriverControls/make.sh
```diff
@@ -0,0 +1,11 @@
...
+    (cd Tests/ && make clean)
+    exit
+fi
+
+(cd ../ && ./format.sh EpsilonDriverControls)
+(make -j4 && cd ../EpsilonDriverControls/ && make)
```
bill-luu, line 10: Same here, moving `../` when formatting doesn't change where you are because that is within parenthesis, so you don't need to `cd` anywhere

JessieG123: Ohhh ok I think I missed read something

bill-luu: lool wait, this line

1. makes the driver controls code
2. cd's out and back into the driver controls code
3. makes the driver controls code again

Then the next line makes the driver controls again before making the test

PR #171, opened 2020-10-04
SFT-198 Trips Unit Tests

duan-le:


EpsilonAuxBmsV2/Tests/tests/TripTest/TripTest.c
```diff
@@ -0,0 +1,134 @@
...
+    RUN_TEST(test_checkChargeTripDueToPackCurrent);
+    RUN_TEST(test_checkChargeTripForNoTrip);
+    RUN_TEST(test_checkProtectionTripForTrip);
+    RUN_TEST(test_checkProtectionTripForNoTrip);
+}
+
+void test_checkDischargeTripDueToLowCell()
+{
+    OrionCanInfo message = {0};
+    AuxTrip auxTripToUpdate = {0};
```
badeniran, line 23: I would suggest making these global variables and only setting the `message` as necessary. This will make sure that state carried between the tests doesn't affect things

badeniran (changes requested):
Looks good! Just the moving of the aux trip and message variables to be global and running `./format.sh`

badeniran (approved):
LGTM. Get one more approval, then squash your commits, then go ahead and merge

JessieG123 (commented):
Also I think including messages would be beneficial so that if it does fail it would be easier to find out as this is probably the purpose of unit testing.

badeniran:
> Would including a short message would be beneficial for the ones that have multiple tests so that if it does fail it would be easier to find out?

Yeah, this is definitely a good idea

JessieG123 (approved):
LGTM!

PR #177, opened 2020-11-08
User/embedded/sft 200 create functions relating to can tx work

badeniran:


badeniran:
So I think we messed up a little bit with our branching, so this is the most I could squash without it bringing up more conflicts to resolve haha

islamrahat735 (approved):

duan-le (approved):

PR #178, opened 2020-11-27
User/embedded/sft 202 creating functions and unit tests relating to aux voltage reading work

badeniran:


duan-le (approved):

JessieG123 (approved):

badeniran:
Not squashing because of bad decisions made when branching :\

PR #190, opened 2022-03-26
SFT-318 Swapped M0 with M1

Hesham-E:
Swapped the #define macros for each motor with each other since the
motor controllers in the car are currently swapped as well. If the
controllers are swapped to their original configuration, then these
changes need to be reverted.


JessieG123 (commented):
It looks like it could work but I would want to test this change before merging it to master. Testing to be done as soon as the AuxBMS and Pis are good.

PR #195, opened 2022-06-03
SFT-327 Trip if you don’t have messages from the orion after 400ms

brian-ngyn:


JessieG123 (approved):
sealofapproval

MarceloLiGonzales (approved):
10/10 would review again

